PR #6, opened 2026-01-07
use public address for ble pairing for consistent address between pairing and reconnection

matttt:
I was having issues with consistency with the address my host was sending to the client, this seems to have fixed that in my case.

May have to do with the fact that I am using bumble's built in `cryptography` package fallbacks (couldnt get cryptography module working on my kindle) 

image

### DIFF
--- a/kindle_hid_passthrough/host.py
+++ b/kindle_hid_passthrough/host.py
@@ -16,7 +16,7 @@ import logging
 from typing import Optional, List, Dict
 
 from bumble.device import Device, Peer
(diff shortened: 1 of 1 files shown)


zampierilucas:
This works, and I also find the same issue with classic devices and fixed it in here  based on your pr https://github.com/zampierilucas/kindle-hid-passthrough/commit/b51cbbe56e3adbfd19fbb005c07ec61afc2a99a3.
Thanks

PR #65, opened 2026-06-11
fix(ble): reconnect to bonded devices that rotate their address

### LINKED ISSUE #62: BLE device does not auto-reconnect after reboot
Hi,
I have a jailbroken Kindle 11 running firmware 5.18 with kindle-hid-passthrough v3.4.1.
I paired a BLE device successfully via BTManager. It works fine when connected.

Problem: After a reboot or after the daemon restarts, the device never auto-reconnects. 

The log shows:
Waiting for connection (Classic: 0, BLE: 1)
[BLE] Waiting for 1 device(s) (accept list)
Connection timeout - no device connected
The only way to reconnect is to manually scan from BTManager every time.

Is there a workaround or fix planned for BLE auto-reconnect?
Thank you!

zampierilucas:
Fixes #62

BLE auto-reconnect waited on the controller filter accept list with the addresses stored in devices.conf. Devices that rotate their address never match, so reconnect timed out forever and every manual reconnect added a duplicate devices.conf entry.

## What changed

Two layers, matching how mainstream stacks (BlueZ/Android/Zephyr) handle BLE Privacy:

1. **Controller address resolution (the standard path).** When the controller supports LL Privacy (the MT8112 does), the resolving list is programmed from the keystore IRKs and address resolution is enabled, so the accept list matches rotating RPAs in hardware with zero added latency. Bonded peers are put in device privacy mode so identity-address advertising isn't filtered (the same issue BlueZ patched in 2021). Accept-list entries now include keystore identity addresses and carry the address type recorded at pairing (they were always added as RANDOM before, which silently broke public-address devices).

2. **Scan fallback.** For controllers without LL Privacy, and for non-compliant devices like the reporter's IINE Gamebrick (rotates its "public" identity and distributes a junk IRK each pairing, so no IRK can ever resolve it), the wait now alternates: accept-list initiating window (18s), short legacy scan (8s). The scan matches by known address, host-side IRK resolution, or configured device name as last resort, then connects directly and re-pairs if needed. After a name-match re-pair the new address is saved and stale same-name entries plus their dead keys are pruned. No standard stack auto-reconnects such devices at all; the name match is deliberately a last-resort fallback.

Also fixed along the way: Classic page-timeout failure events killed the BLE wait (missing transport filter), and the "Future exception was never retrieved" log spam after connection-cancel.

## Controller quirks (MT8112, Kindle Basic 5)

- Cannot scan while in the initiating state (COMMAND_DISALLOWED), hence alternating windows.
- Refuses extended LE commands once a legacy one was issued in the session, so the reconnect path is all-legacy.
- Stays in the initiating state until the cancel's connection-complete event arrives; the cancel waits for it before scanning.

## Testing

- Probed the MT8112: LL Privacy feature bit set, all resolving-list commands supported, 64-entry accept list.
- On-device: resolving list + privacy mode programming succeeds, clean wait cycles with zero HCI errors, Classic handler unaffected.
- Matcher logic verified on-device with real bumble types: known address, name match from a rotated address, IRK/RPA resolution with a real ah() hash, and negative cases.
- End-to-end rotated reconnect needs a physical rotating device; asking the reporter to verify with the Gamebrick.

### DIFF
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -51,9 +51,11 @@ class BLEMixin:
         await self.device.send_command(
             HCI_LE_Clear_Filter_Accept_List_Command(), check_result=True)
 
-        for addr_str in addresses:
+        known = {normalize_addr(a) for a in addresses} | self._keystore_addresses
+
+        for addr_str in sorted(known):
             target = Address(addr_str)
-            known_type = self._keystore_address_types.get(normalize_addr(addr_str))
+            known_type = self._keystore_address_types.get(addr_str)
             if known_type is not None:
                 entry_types = [known_type & 1]
             else:
@@ -66,8 +68,6 @@ class BLEMixin:
                     ), check_result=True)
 
         log.info(f"[BLE] Waiting for {len(addresses)} device(s) (accept list)")
-
-        known = {normalize_addr(a) for a in addresses} | self._keystore_addresses
 
         try:
             connection = None
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -3,7 +3,7 @@
 
 import asyncio
 
-from bumble.core import BT_LE_TRANSPORT, InvalidStateError
+from bumble.core import AdvertisingData, BT_LE_TRANSPORT, InvalidStateError
 from bumble.device import Device, Peer
 from bumble.gatt import (
     GATT_DEVICE_NAME_CHARACTERISTIC,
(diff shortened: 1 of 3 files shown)


zampierilucas:
@Masaju31 could you give this a try? Grab the `kindle-hid-passthrough-armv7` artifact from the CI build on this PR and install it like a release tarball, re-pair the Gamebrick once via BTManager, then power cycle it a few times and reboot the Kindle. If it still won't auto-reconnect, paste the log from `/var/log/hid_passthrough.log`. thanks!

Masaju31:
Thank you so much, it works perfectly !
After re-pairing, the Gamebrick auto-reconnects every time after a power cycle and after rebooting the Kindle.
I'm so happy to use my kindle like this, thank you again for your work :)

zampierilucas:
@Masaju31 awesome merging then, should be available on the next version from kindle forge!

PR #69, opened 2026-06-28
fix: ignore malformed BLE advertisements

kal-kaliper:
## Summary
- ignore malformed BLE local-name advertising data during rotated-address matching
- ignore malformed advertising records during BLE HID scanning instead of aborting the scan callback

## Context
On a Kindle PW5 running `3.6.0-5e99c17`, a paired BLE keyboard failed to reconnect after sleep. The daemon reached `[BLE] Waiting for 1 device(s) (accept list)`, then the rotated-address scan repeatedly failed with:

```text
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
  File "/mnt/us/kindle_hid_passthrough/dist/ble.py", line 240, in _match_rotated_ble_device
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/core.py", line 2094, in get
```

This makes reconnect fragile when any nearby BLE advertiser has malformed local-name data. The bad advertisement should be skipped, not allowed to break reconnect or scanning.

## Test
- `python3 -m py_compile kindle_hid_passthrough/ble.py kindle_hid_passthrough/scanner.py`

### DIFF
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -236,8 +236,12 @@ class BLEMixin:
                              f"{self._format_device(resolved_norm)}")
                     return (dev, 'irk')
 
-        name = advertisement.data.get(AdvertisingData.COMPLETE_LOCAL_NAME) or \
-            advertisement.data.get(AdvertisingData.SHORTENED_LOCAL_NAME)
+        try:
+            name = advertisement.data.get(AdvertisingData.COMPLETE_LOCAL_NAME) or \
+                advertisement.data.get(AdvertisingData.SHORTENED_LOCAL_NAME)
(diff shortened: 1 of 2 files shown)


kal-kaliper:
Marking this as draft pending on-device validation. The patch is based on observed Kindle logs from a PW5 where malformed advertisement data repeatedly raised `UnicodeDecodeError` during rotated-address reconnect, but I have not yet rebuilt and deployed the ARM binary to confirm the full reconnect path is fixed.

kal-kaliper:
Closing this for now. I opened it before completing on-device validation, so I am going to validate the fix on hardware first and come back with a confirmed patch if it works.

PR #70, opened 2026-06-28
fix: ignore malformed BLE advertisements

kal-kaliper:
## What

Ignore malformed BLE advertising records instead of letting Bumble's advertising data decoder abort scan or reconnect handling with a `UnicodeDecodeError`.

The fix is intentionally narrow:

- during paired BLE reconnect, skip a malformed advertisement when matching rotated BLE addresses by advertised name
- during BTManager BLE scanning, skip malformed HID-service advertisements or fall back to `Unknown` when only the local name is malformed

## Practical problem

On a Kindle using BTManager for a BLE keyboard, a nearby device can advertise malformed local-name data. When Bumble decodes that advertising record, it may raise `UnicodeDecodeError`. Before this change, that exception could break the BLE scan/reconnect path even though the paired keyboard itself was present and bonded.

In practice this showed up as BTManager failing to find or reconnect a paired Bluetooth keyboard: the UI and daemon were running, the keyboard was nearby, but the reconnect loop could be interrupted by an unrelated malformed BLE advertisement in the environment.

## Validation

Tested on a Kindle PW5 with a paired BLE keyboard.

- built a validation runtime containing this change via GitHub Actions on a fork
- installed the runtime on the Kindle while preserving existing pairing keys and BTManager config
- confirmed the daemon reconnected to the paired keyboard and exposed `/dev/input/event2`
- confirmed recent daemon logs no longer showed the previous `UnicodeDecodeError` from BLE advertisement parsing after the patched daemon started
- confirmed BTManager `/status` reported the keyboard connected with HID descriptor and input path
- ran a sleep/wake validation cycle where the Kindle reached `readyToSuspend`, returned to `active`, and the keyboard remained connected afterward

Local check:

- `python3 -m py_compile kindle_hid_passthrough/*.py`

## Notes

This PR only addresses malformed BLE advertisements interrupting scan/reconnect. It deliberately does not include separate follow-up work for stale controller-side BLE links after suspend or BTManager close behavior.

### DIFF
--- a/kindle_hid_passthrough/scanner.py
+++ b/kindle_hid_passthrough/scanner.py
@@ -141,9 +141,6 @@ class Scanner:
             if addr_str in seen_addresses:
                 return
 
-            # Check for HID service in advertising data. Some nearby devices
-            # advertise malformed local names, and Bumble may decode them while
-            # iterating the advertising records.
             is_hid = False
             if hasattr(advertisement, 'data') and advertisement.data:
                 try:
--- a/kindle_hid_passthrough/scanner.py
+++ b/kindle_hid_passthrough/scanner.py
@@ -141,24 +141,30 @@ class Scanner:
             if addr_str in seen_addresses:
                 return
 
-            # Check for HID service in advertising data
+            # Check for HID service in advertising data. Some nearby devices
+            # advertise malformed local names, and Bumble may decode them while
+            # iterating the advertising records.
             is_hid = False
             if hasattr(advertisement, 'data') and advertisement.data:
-                services = advertisement.data.get(
-                    AdvertisingData.COMPLETE_LIST_OF_16_BIT_SERVICE_CLASS_UUIDS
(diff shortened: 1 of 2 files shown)


kindle_hid_passthrough/scanner.py
```diff
@@ -141,33 +141,43 @@ def on_advertisement(advertisement):
             if addr_str in seen_addresses:
                 return
 
-            # Check for HID service in advertising data
+            # Check for HID service in advertising data. Some nearby devices
+            # advertise malformed local names, and Bumble may decode them while
+            # iterating the advertising records.
```
zampierilucas: dont care get rid of this change

zampierilucas:
Other than the small nit, LGMT, thanks for the pr!

kal-kaliper:
Addressed the review nit by removing the explanatory scanner comment. No behavior changes.

zampierilucas (approved):

PR #71, opened 2026-06-28
fix: close BTManager without killing mesquite

kal-kaliper:
## What

Close BTManager through Kindle app manager lifecycle instead of asking the helper API to force-kill the Mesquite WAF process.

The close button now prefers `kindle.appmgr.back()` and keeps `startApplication("com.lab126.booklet.home")` as a fallback. The `/quit-app` request and backend `pkill mesquite.*BTManager` handler are removed.

## Practical problem

The previous `/quit-app` workaround was added to avoid a ~19 MB Mesquite RSS leak after closing BTManager. In practice it also terminated Mesquite from outside appmgrd, which can make Kindle treat BTManager as if it crashed and show an application error after close.

This keeps the memory-cleanup intent while avoiding the crash-looking forced termination path.

## Validation

Tested on a Kindle PW5 with BTManager installed.

- installed a device-only version of this close behavior
- confirmed the installed UI had no `/quit-app` or `pkill` reference
- closed BTManager from the UI and confirmed no BTManager Mesquite process remained afterward, so the RSS cleanup still happened
- confirmed the HID daemon stayed running and the paired keyboard remained connected on `/dev/input/event2`
- checked current Kindle logs after close and found no BTManager `application_state_error`, `abnormal_exit`, or Mesquite close error for the installed variant

Earlier validation of the no-`/quit-app` path also showed appmgrd reclaiming Mesquite normally with `exitcode=0` and `abnormal_exit=0`.

Local check:

- `python3 -m py_compile kindle_hid_passthrough/*.py`

## Notes

This PR changes only BTManager close behavior. It does not change pairing, scanning, reconnect behavior, or daemon lifecycle.

### DIFF
--- a/illusion/BTManager/script.js
+++ b/illusion/BTManager/script.js
@@ -647,12 +647,11 @@ var BTManager = (function() {
 
     function quit() {
         pressBtn("btnBack");
-        if (typeof kindle !== "undefined" && kindle.appmgr && kindle.appmgr.startApplication) {
-            kindle.appmgr.startApplication("com.lab126.booklet.home");
-        } else if (typeof kindle !== "undefined" && kindle.appmgr && kindle.appmgr.back) {
+        if (typeof kindle !== "undefined" && kindle.appmgr && kindle.appmgr.back) {
             kindle.appmgr.back();
+        } else if (typeof kindle !== "undefined" && kindle.appmgr && kindle.appmgr.startApplication) {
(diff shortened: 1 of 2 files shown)


zampierilucas:
HI @kal-kaliper, I've tested your code and this does not work, as the mesquite child keeps running indeterminately in the background uses memory, and more importantly leaking memory, this will inevitably lead to the kindle ooming and rebooting. 
Dont open a pr of untested code again.

PR #81, opened 2026-07-01
Simplified installation

kbarni:
Here's a suggestion for a simplified installation procedure: the prepared tarball contains the folders to be copied to the root of the Kindle, so the booklet, the kual extension, the koreader plugin and the binaries are installed to the correct destination.

The only install action left for the user are the copying of the udev rules and the upstart service (which need a script to mount the root partition as rw). The other install actions were removed.

Check the resulting file here: https://github.com/kbarni/kindle-hid-passthrough/actions/runs/28549142609

What do you think?


zampierilucas:
This is indeed, much better, I just want to be sure we wont break the kindleForge install https://github.com/KindleTweaks/Repository/tree/main/KindleHIDPassthrough before merging this

PR #89, opened 2026-07-04
Stabilize mixed BLE and Classic HID lifecycle

kal-kaliper:
## Summary

- Finalizes the mixed BLE and Classic HID session lifecycle so each protocol owns its own session state and reconnect path.
- Adds Classic phone flap handling: short remote drops without input now back off active dials instead of continuously churning UHID devices.
- Quietens the worst background churn by sleeping through all-device Classic flap backoff and disabling Classic Page Scan during those backoff windows.
- Tightens Kindle power lifecycle handling by suspending on `goingToScreenSaver` and making the power event monitor idempotent.
- Preserves BLE keyboard behavior and the cached Classic descriptor fallback used by Pixel HID apps.

## Validation

- `python3 -m unittest discover -s tests`
- GitHub ARM build succeeded on fork branch: https://github.com/kal-kaliper/kindle-hid-passthrough/actions/runs/28699476934
- Deployed that artifact to a PW5 Kindle and confirmed API status reports `version: 3.6.0-5d66f32`.
- Kindle smoke logs show the Pixel still drops with HCI reason 19 when the phone-side HID app is inactive, but retry attempts now respect escalating backoff and Page Scan is disabled during the quiet part of the backoff window.

## Log audit notes

The original Kindle logs showed a storm of `conn2ap_btif0_wakeup_out_b` wake IRQs, repeated Pixel Classic connect/drop cycles, KOReader input detach churn from transient UHID devices, and stale/noisy power monitor processes. This branch addresses the daemon-owned parts of that pattern.

A remaining caveat: `conn2ap_btif0_wakeup_out_b` still appears in `dmesg` after deployment, including from WMT/kernel contexts. The app now reduces active Classic scanning during flap backoff, but the remaining IRQ activity looks lower-level than the daemon retry loop and should be treated as a follow-up hardware/WMT investigation rather than hidden as fixed here.


kal-kaliper:
Closing pending owner approval and PR split/rework. This was opened prematurely as a draft integration PR.

PR #95, opened 2026-07-04
KO3/KPW4 recovery fixes on 98b9e5b base

rubyxs:
I forked from commit `98b9e5b` (`fix(uhid): backport UHID_CREATE2/INPUT2 into heisenberg module`) because I couldn’t get `main` / `v3.8.0` working reliably on my Oasis 3. This older base got much closer to working on both my Kindle Paperwhite 4 and Oasis 3 after the following local changes, although there are still a few bugs.

The changes were concentrated in four files:

- `scripts/hid-passthrough-daemon.sh`
- `scripts/run-daemon-pw4.sh`
- `illusion/BTManager.sh`
- `koreader-plugin/hidpassthrough.koplugin/main.lua`

The intent of saving this fork is not to claim “this is the final solution,” but rather: this was the last base that worked for me on KPW4 and got closest on KO3, so maybe there is something in these local changes that is still useful to bring back into `main`.

What those changes were trying to address:

- more reliable KO3/PW4 startup and `uhid` bring-up
- recovery from `api_only` / half-started daemon states
- recovery when the daemon is up but usable keyboard input is still missing
- a few on-device recovery actions in the KOReader plugin, especially `Restart daemon wrapper` and `Radio reset for BLE`

### DIFF
--- a/illusion/BTManager.sh
+++ b/illusion/BTManager.sh
@@ -10,11 +10,16 @@ APP_DIR="/mnt/us/kindle_hid_passthrough/illusion/BTManager"
 BINARY="/mnt/us/kindle_hid_passthrough/kindle-hid-passthrough"
 LD_PROCESS="ld-linux-armhf."
 APPREG_DB="/var/local/appreg.db"
+RUNNER="/mnt/us/kindle_hid_passthrough/scripts/run-daemon-pw4.sh"
 
 # ---- Helpers ----
 
 log_msg() {
     logger -t btmanager "$1"
+}
+
+has_uhid() {
+    [ -e /dev/uhid ]
 }
 
 alert() {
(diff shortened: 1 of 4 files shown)


zampierilucas:
Hey @rubyxs, thanks for writing this up instead of just sitting on the fork, it made it easy to see what you were actually hitting.

You forked at 98b9e5b, which is two commits before 44835c3, and that one is almost certainly your whole problem. Under the nuitka build the daemon resolved its module dir from `__file__`, which doesn't point at `dist/kindle_hid_passthrough/modules`, so it logged "no bundled uhid.ko" while the module was sitting right there. Exporting `KINDLE_HID_BASE` and insmodding by hand in `run-daemon-pw4.sh` is exactly the workaround for that, so you were papering over something that got fixed a couple of days after you branched.

Most of the rest has a proper fix now too. 51c9ed3 retries `ensure_uhid()` on reconnect and adds `hid_ready` to `/status`, which is your "daemon up but no usable input" check done at the right layer. 384bebe closes the sockets the daemon inherits from KOReader, that was causing the half started api_only state when you launched from the plugin. And 961183c is the real version of your radio reset, on zelda the BCM chip can still be asleep when bumble sends the first HCI command, so it now asserts wake and confirms a raw HCI Reset before handing the chip over. Your wifi toggle worked because it bounces the same combo chip.

Could you grab v3.11.0 and give it another go on the PW4 and the O3?
https://github.com/zampierilucas/kindle-hid-passthrough/releases/tag/v3.11.0

The one bit in here that's still genuinely broken on main is `stop()` in hid-passthrough-daemon.sh, pgrep can return more than one pid and then the kill just errors out and nothing stops. I'll pull that in separately.

Issue #75, opened 2026-07-01
Bug Report: HCI Reset timeout during reconnection after device disconnection


ZongbingHou:
# Bug Report: HCI Reset timeout during reconnection after device disconnection

**Build:** https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/28479280603
**Device:** Kindle Oasis 3 (code 0x3D4)
**Bluetooth Chip:** BCM (Broadcom)
**UART:** /dev/ttymxc2

## Description

After successfully pairing and connecting to a BLE HID device (IINE_keyboard), when the device disconnects, the daemon attempts to reconnect. During the reconnection process, `HCI Reset` command consistently times out, preventing reconnection.

Additionally, after the daemon enters a failed state, subsequent scan attempts also exhibit `HCI Reset` timeout issues. The daemon's suspend/resume cycle appears to interfere with reconnection attempts.

## Steps to Reproduce

1. Pair a BLE HID device (e.g., IINE_keyboard) with Kindle Oasis 3 via `kindle_hid_passthrough`.
2. Allow the device to connect successfully and receive HID reports.
3. Let the device disconnect (either by distance, power off, or automatic disconnection).
4. Observe the daemon's automatic reconnection attempts.

## Observed Behavior

- The initial pairing and connection succeed.
- After disconnection, the daemon attempts to reconnect.
- Each reconnection attempt sends `HCI Reset`, which times out after 10 seconds.
- The daemon retries indefinitely, failing every time.
- In some cases, `Daemon suspending...` is logged during the timeout period.
- After a successful reconnection (rare), the connection eventually fails again with `HCI_DISCONNECT_COMMAND timed out`.
- After disconnection, the daemon may report "No devices configured, waiting for pairing..." even though a device was previously paired.

### 1. Initial pairing and connection (successful)

```
2026-07-01 10:09:35,265 INFO ble_hid: Devices: 0 Classic, 0 BLE
2026-07-01 10:09:35,267 INFO ble_hid: HID Host v3.7.0-cecdc3b
2026-07-01 10:09:36,285 INFO ble_hid: Sending HCI Reset...
2026-07-01 10:09:36,326 INFO ble_hid: HCI Reset successful
2026-07-01 10:09:36,580 INFO ble_hid: Device powered on: AC:CC:FC:3C:44:DC/P
2026-07-01 10:09:36,581 INFO ble_hid: BLE enabled
[...]

PR #98, opened 2026-07-05
fix(koplugin): keep keyboard attached across FileManager/ReaderUI switches

zampierilucas:
Follow-up to #75: [ZongbingHou reported](https://github.com/zampierilucas/kindle-hid-passthrough/issues/75#issuecomment-4884712788) that every book open/close bounces the keyboard, disconnect/reconnect toasts and a few seconds of dead keyboard, even though the BT link never drops.

The daemon is innocent this time. KOReader tears down and recreates plugin instances on every FileManager to ReaderUI switch, and our onCloseWidget assumed it only ran at KOReader exit: it detached every keyboard (closed the input fd, restored device caps, broadcast the disconnect event). The new instance then waited out the 2s HTTP probe plus the watcher tick before reattaching, so the keyboard was dead for 3 to 6 seconds with two toasts on every open/close.

What's in here:

1. Attachment state (kb_attached, kb_count, kb_original_caps) moved to module locals so it survives instance churn. The old class/instance mix only stayed consistent because detach always ran before teardown.
2. onCloseWidget now stops only this instance's polling and cancels the pending init probe. The fds stay registered in the global Device.input, so the keyboard keeps working across the switch.
3. Detach logic split into _detachAllKeyboards, still called on daemon stop, and the reconciler still detaches naturally when the uhid device disappears.
4. init() re-arms the watcher immediately when keyboards are already attached instead of waiting for the 2s probe.

### DIFF
--- a/koreader-plugin/hidpassthrough.koplugin/main.lua
+++ b/koreader-plugin/hidpassthrough.koplugin/main.lua
@@ -215,12 +215,10 @@ end
 local function yes() return true end
 local function no()  return false end
 
--- Map of attached keyboard event paths -> { fd_object, original_caps_index }.
--- The fd_object is whatever Device.input:fdopen returns, which we hand back
--- to :close() on removal.
-HIDPassthrough._kb_attached = {}
-HIDPassthrough._kb_count = 0
-HIDPassthrough._kb_original_caps = nil
+-- Module-level so attachments survive FileManager <-> ReaderUI re-instantiation.
+local kb_attached = {}
+local kb_count = 0
+local kb_original_caps = nil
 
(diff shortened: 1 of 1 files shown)


zampierilucas:
Hey @ZongbingHou if you have a change, this should make the plugin connection less flaky, and improve the reconnects

ZongbingHou:
It's effective, now when I open and close the book, it no longer repeatedly shows disconnecting and reconnecting.

mergen3107:
Sorry I am so lost with all those revision :D 
Can I just wait until everything gets merged to test it?

zampierilucas:
Thank you @ZongbingHou !
And @mergen3107 yeah, so do I, merging this now, and cutting a new release later today.

PR #99, opened 2026-07-05
fix(bt): clean stock BT handoff on MTK takeover to stop Scribe boot loop (#87)

### LINKED ISSUE #87: Boot loop
## Title

Kindle Scribe 1 on firmware 5.17.3 boot-loops and loses network after enabling HID Passthrough

## Description

Enabling `kindle-hid-passthrough` autostart on my first-generation Kindle Scribe causes the Kindle to enter a boot loop or boot-like restart cycle. During the failure, network connectivity also disappears, including SSH access.

I can briefly connect over SSH during the boot window. Disabling the `hid-passthrough` Upstart job allows the Kindle to boot and remain stable again.

The 8BitDo Zero 2 itself pairs and works in keyboard mode. The problem appears to occur when HID Passthrough starts automatically and takes control of the MediaTek Bluetooth device during Kindle startup.

## Environment

* Device: Kindle Scribe, first generation
* Kindle firmware: 5.17.3
* Jailbroken: yes
* Bluetooth hardware path: `/dev/stpbt`
* Controller: 8BitDo Zero 2
* Controller mode: keyboard mode
* HID protocol: Classic Bluetooth, as far as I understand
* KOReader: older version intentionally retained because the latest KOReader version breaks pen functionality on this firmware/device
* `kindle-hid-passthrough` version: 3.8.0
* Installation method: manual install.sh - selected 1. everything

Related but different Zero 2 issue: #85.

## Expected behavior

HID Passthrough should start without affecting the Kindle’s ability to finish booting or retain Wi-Fi/USB networking.

The 8BitDo Zero 2 should reconnect and expose a UHID/input device that can be used for page-turn actions.

## Actual behavior

After installing/enabling HID Passthrough autostart:

1. The Kindle starts booting.
2. SSH becomes available briefly at `192.168.15.244`.
3. HID Passthrough appears to start during this window.
4. The Kindle then loses network connectivity.
5. The device returns to the boot screen or otherwise restarts the startup sequence.
6. The behavior repeats.

Because SSH is only available briefly, I initially could not disable the daemon before the connection disappeared.

Removing or renaming `/etc/upstart/hid-passthrough.conf` during the short SSH window stopped the loop and returned the Kindle to a stable state.

## Recovery performed

I recovered the Kindle with a direct SSH command that:

```sh
stop hid-passthrough 2>/dev/null || true
pkill -9 -f '[k]indle[-_]hid[-_]passthrough' 2>/dev/null || true
mntroot rw

mv /etc/upstart/hid-passthrough.conf \
   /etc/upstart/hid-passthrough.conf.disabled

sync
reboot -f
```

After disabling that Upstart job, the Kindle boots normally and network access remains available.

## Upstart/service information

The relevant Kindle Upstart relationships on this Scribe are:

```text
/etc/upstart/acsbtfd.conf:1:start on started btmanagerd
/etc/upstart/acsbtfd.conf:15:su bluetooth -c "exec /usr/bin/acsbtfd -f 2>> $STDERR"

/etc/upstart/btmanagerd.conf:7:env STDERR=/tmp/btmanagerd.err
/etc/upstart/btmanagerd.conf:19:# pre-condition for btmanagerd to start.
/etc/upstart/btmanagerd.conf:26:    if [ "$btmanagerd" = "1" ]; then
/etc/upstart/btmanagerd.conf:48:      exec nice -n -10 btmanagerd -f 2>> $STDERR

/etc/upstart/wmt.conf:1:start on (starting acsbtfd or starting wifim)
/etc/upstart/wmt.conf:7:    exec wmt_service -- -p /lib/firmware
```

This may be relevant because HID Passthrough starts on `started wmt`, while the Kindle’s `wmt` service is connected to both the Bluetooth and Wi-Fi startup paths.

## `/dev/stpbt` holder check

With the Kindle recovered, stable, and HID Passthrough disabled, I checked all `/proc/*/fd` links for a process holding `/dev/stpbt`.

Command:

```sh
for p in /proc/[0-9]*; do
    pid=${p##*/}

    for fd in "$p"/fd/*; do
        target=$(readlink "$fd" 2>/dev/null)

        if [ "$target" = "/dev/stpbt" ]; then
            echo "PID: $pid"
            echo "COMM: $(cat "$p/comm" 2>/dev/null)"
            echo "EXE:  $(readlink "$p/exe" 2>/dev/null)"
            echo "CMD:  $(tr '\000' ' ' < "$p/cmdline" 2>/dev/null)"
            echo "FD:   $fd"
            echo
            break
        fi
    done
done
```

Result:

```text
No holder was found.
```

This was only a point-in-time check after recovery. It does not rule out a process opening `/dev/stpbt` transiently during boot.

## Missing application log

After recovery:

```sh
cat /var/log/hid_passthrough.log
```

returned:

```text
cat: can't open '/var/log/hid_passthrough.log': No such file or directory
```

The Upstart configuration redirects the program’s stdout and stderr to `/dev/null`, so there may be no persistent evidence if the process fails before its file logger is initialized.

It would be useful for the Upstart job to preserve early startup output, for example in:

```text
/mnt/us/kindle_hid_passthrough/startup.log
```

or `/var/log/hid_passthrough-startup.log`.

## Suspected area

This is only a hypothesis, not yet confirmed.

On MediaTek devices, the current startup code tries to free `/dev/stpbt` and may invoke:

```sh
fuser -k /dev/stpbt
```

It can also scan `/proc` and send `SIGKILL` to processes holding the node.

Because the Kindle’s `wmt` startup is associated with both `acsbtfd` and `wifim`, it may be unsafe to indiscriminately kill an unknown holder while those connectivity services are still starting.

Possible failure sequence:

1. `wmt` reports started.
2. `hid-passthrough` starts immediately.
3. A Kindle Bluetooth/connectivity process briefly owns `/dev/stpbt`.
4. HID Passthrough kills or races with that process.
5. The Kindle connectivity stack or a watchdog restarts.
6. Upstart respawns HID Passthrough, repeating the conflict.

I do not yet have logs proving this sequence.

## Suggested diagnostic improvements

Would you consider adding the following before killing a `/dev/stpbt` holder?

* Log the holder PID.
* Log `/proc/<pid>/comm`.
* Log `/proc/<pid>/exe`.
* Log `/proc/<pid>/cmdline`.
* Log the UID/GID.
* Log whether the PID belongs to a known Kindle service.
* Send `SIGTERM` first and wait before using `SIGKILL`.
* Refuse to kill unknown processes unless an explicit force option is enabled.
* Preserve early stdout/stderr instead of redirecting it to `/dev/null`.
* Add a MediaTek/Scribe startup delay or wait for an explicit readiness condition.
* Avoid an Upstart respawn loop if hardware preparation fails.
* Restore the stock Bluetooth owner cleanly when HID Passthrough exits.

It may also be safer for autostart to require multiple conditions instead of starting immediately on:

```text
started wmt
```

## Additional logs I am collecting

I am preparing a manual, non-Upstart reproduction with:

* `/var/log/messages`
* repeated `dmesg` snapshots
* boot ID and uptime
* Upstart job status
* `/dev/stpbt` holder changes
* `btmanagerd`, `acsbtfd`, `wmt_service`, `wifid`, and HID process state
* network interface state
* daemon stdout/stderr
* `/var/log/hid_passthrough.log`, when created
* `/proc/last_kmsg` and `/sys/fs/pstore`, if available

I will attach those logs once I can reproduce the failure with autostart disabled and an automatic safety timeout.

## Questions

1. Is the first-generation Kindle Scribe on firmware 5.17.3 known to work with the current MediaTek takeover implementation?
2. Is killing every process holding `/dev/stpbt` expected on a Scribe?
3. Which stock process is normally expected to own `/dev/stpbt` immediately after `wmt` starts?
4. Could starting on `started wmt` be too early on this firmware?
5. Is there a debug flag or unbundled Python invocation that would provide more verbose Bumble/HCI logging?
6. What is the preferred command for reporting the exact installed HID Passthrough version when the daemon cannot stay running?
7. Would a PR adding holder identification, safer termination, and persistent early-startup logging be welcome?

I am willing to test patches and prepare a PR once the failure point is identified.

zampierilucas:
@cosbot1 reported a boot loop on a Scribe (5.17.3) in #87, enabling hid-passthrough autostart makes the kindle boot loop and lose wifi/ssh, and disabling the upstart job brings it back. @mergen3107 reproduced it on a KS1 and ground through every iteration of this, huge thanks.

The root cause turned out to be the start trigger, not the eviction. `start on started wmt` fires the instant wifi begins bring-up (wmt itself starts on `starting wifim`), so we were loading the BT module and opening /dev/stpbt while the CONSYS combo chip was still initializing. mergen's last_kmsg shows mtk_wmtd still running SoC calibration 30s into boot, which is why his `sleep 30` was the only thing that ever worked and a 10-12s trigger (lab126_gui) still looped.

What's in here:

1. the job now starts on `framework_ready`, the event the framework emits right after tearing down the boot splash. it's the same gate amazon uses for its own deferrable services, so no magic number and no collision window. the keyboard comes up with the home screen.
2. a boot-loop breaker in pre-start. starts are counted in `boot_attempts` on /mnt/us (which survives reboots, unlike /var), after 3 starts without a stable run the job disables itself, and every start snapshots /proc/last_kmsg plus a line in `boot.log`. worst case a bad build loops 3 times, recovers on its own, and leaves a postmortem. the daemon clears the counter after 2 minutes alive, enabling autostart resets it.
3. when /dev/stpbt is actually held, prepare() identifies each holder and stops the stock BT stack through upstart (`initctl stop`) instead of SIGKILL, so it doesn't respawn and fight us. shared wifi/CONSYS processes are never touched, we bail instead of risking wifi. anything else gets SIGTERM first.
4. the release upstart job logs to /var/log/hid_passthrough.log instead of /dev/null.

Tested by @mergen3107 on his KS1: clean boot on the first try, breaker and postmortem files all working. Fixes #87.

### DIFF
--- a/assets/hid-passthrough.upstart
+++ b/assets/hid-passthrough.upstart
@@ -11,5 +11,5 @@ env KINDLE_HID_BASE=/mnt/us/kindle_hid_passthrough
 console none
 
 script
-    exec /mnt/us/kindle_hid_passthrough/kindle-hid-passthrough --daemon >/dev/null 2>&1
+    exec /mnt/us/kindle_hid_passthrough/kindle-hid-passthrough --daemon >>/var/log/hid_passthrough.log 2>&1
 end script
--- a/assets/hid-passthrough.upstart
+++ b/assets/hid-passthrough.upstart
@@ -1,6 +1,6 @@
 # HID Passthrough - Bluetooth HID host with UHID passthrough
 # For use with pre-built binary from releases
-start on started wmt
+start on framework_ready
 stop on stopping filesystems
 
 respawn
@@ -9,6 +9,20 @@ respawn limit 5 60
 env KINDLE_HID_BASE=/mnt/us/kindle_hid_passthrough
 
 console none
+
+pre-start script
+    base=/mnt/us/kindle_hid_passthrough
+    cp /proc/last_kmsg "$base/last_kmsg.txt" 2>/dev/null || true
+    n=$(cat "$base/boot_attempts" 2>/dev/null || echo 0)
+    case "$n" in ''|*[!0-9]*) n=0 ;; esac
+    if [ "$n" -ge 3 ]; then
+        echo "$(date) autostart disabled after $n starts without a stable run" >> "$base/boot.log"
+        stop
+        exit 0
+    fi
+    echo $((n + 1)) > "$base/boot_attempts"
+    echo "$(date) start attempt $((n + 1))" >> "$base/boot.log"
+end script
 
 script
     exec /mnt/us/kindle_hid_passthrough/kindle-hid-passthrough --daemon >>/var/log/hid_passthrough.log 2>&1
--- a/kindle_hid_passthrough/bt_mtk.py
+++ b/kindle_hid_passthrough/bt_mtk.py
@@ -4,6 +4,7 @@
 import glob
 import os
 import signal
+import subprocess
 import time
 
 from bt_chip import BtChip, run
@@ -21,6 +22,10 @@ _STOCK_BT_JOBS = ('acsbtfd', 'btmanagerd')
 # Shared Wi-Fi/CONSYS combo processes — killing one can drop Wi-Fi. Never touch.
 _CRITICAL_COMMS = ('wmt_service', 'mtk_wmtd', 'wifid', 'wifim',
                    'stp_main', 'mtk_stp_psm', 'connfem')
+
+# Upstart rests in these; everything else (pre-start, spawned, stopping, ...) is
+# a transition, and evicting a job mid-transition leaves CONSYS half-configured.
+_SETTLED_STATES = ('running', 'waiting')
 
 
 def _find_bt_module(patterns=None):
@@ -110,6 +115,66 @@ def _find_holders(device_path):
     return holders
 
 
+def _job_state(job):
+    """(goal, state) of an upstart job, or (None, None) if it isn't known here."""
+    try:
+        r = subprocess.run(['/sbin/initctl', 'status', job],
+                           capture_output=True, timeout=5)
+    except Exception:
+        return (None, None)
+    if r.returncode != 0:
+        return (None, None)
+    head = r.stdout.decode(errors='replace').split('\n')[0].split(',')[0]
+    fields = head.split()
+    if not fields or '/' not in fields[-1]:
+        return (None, None)
+    goal, _, state = fields[-1].partition('/')
+    return (goal, state)
+
+
+def _wait_for_settle(jobs, timeout, poll=0.5):
+    """Block until every job leaves its transitional state, or timeout expires.
+
+    Starting on a boot event only says the CONSYS chip powered up, not that the
+    stock BT stack finished claiming it. Evicting acsbtfd mid-init flaps the
+    radio Wi-Fi shares, which is what boot-loops a Scribe (#87).
+    """
+    deadline = time.monotonic() + timeout
+    announced = False
+    while True:
+        unsettled = []
+        for job in jobs:
+            goal, state = _job_state(job)
+            if state is not None and state not in _SETTLED_STATES:
+                unsettled.append(f'{job} {goal}/{state}')
+        if not unsettled:
+            if announced:
+                log.info("Stock BT stack settled")
+            return True
+        if time.monotonic() >= deadline:
+            log.warning(f"Stock BT stack still unsettled after {timeout:.0f}s: "
+                        f"{', '.join(unsettled)}")
+            return False
+        if not announced:
+            log.info(f"Waiting for stock BT stack to settle: {', '.join(unsettled)}")
+            announced = True
+        time.sleep(poll)
+
+
+def _wait_for_stopped(job, timeout, poll=0.5):
+    """Block until an upstart job reaches stop/waiting, or timeout expires."""
+    deadline = time.monotonic() + timeout
+    while True:
+        goal, state = _job_state(job)
+        if state is None or (goal, state) == ('stop', 'waiting'):
+            return True
+        if time.monotonic() >= deadline:
+            log.warning(f"'{job}' did not reach stop/waiting in {timeout:.0f}s "
+                        f"(now {goal}/{state})")
+            return False
+        time.sleep(poll)
+
+
 def _graceful_kill(h, settle, device_path):
     """SIGTERM an unknown holder, then SIGKILL only if it clings to the node."""
     for sig in (signal.SIGTERM, signal.SIGKILL):
@@ -124,7 +189,7 @@ def _graceful_kill(h, settle, device_path):
         time.sleep(settle)
 
 
-def _evict_holders(device_path, settle):
+def _evict_holders(device_path, settle, stop_timeout=10.0):
     """Free device_path: stop stock BT via upstart, spare Wi-Fi, SIGTERM the rest."""
     holders = _find_holders(device_path)
     if not holders:
@@ -145,6 +210,8 @@ def _evict_holders(device_path, settle):
         log.info(f"Stopping stock BT service '{job}' via upstart")
         if not run(['/sbin/initctl', 'stop', job]):
             log.warning(f"'initctl stop {job}' failed")
+    for job in stock_jobs:
+        _wait_for_stopped(job, stop_timeout)
     if stock_jobs:
         time.sleep(settle)
 
@@ -192,8 +259,14 @@ class MtkChip(BtChip):
             log.info(f"{device_path} is available")
             return True
 
+        log.info(f"{device_path} is busy, waiting for the stock BT stack to settle...")
+        _wait_for_settle(_STOCK_BT_JOBS, config.bt_stock_settle_timeout)
+        if _is_device_free(device_path):
+            log.info(f"{device_path} is available")
+            return True
+
         log.info(f"{device_path} is busy, identifying and evicting holders...")
-        _evict_holders(device_path, settle)
+        _evict_holders(device_path, settle, config.bt_stop_timeout)
         if _is_device_free(device_path):
             log.info(f"{device_path} is now available")
             return True
--- a/kindle_hid_passthrough/bt_mtk.py
+++ b/kindle_hid_passthrough/bt_mtk.py
@@ -4,7 +4,6 @@
 import glob
 import os
 import signal
-import subprocess
 import time
 
 from bt_chip import BtChip, run
@@ -22,10 +21,6 @@ _STOCK_BT_JOBS = ('acsbtfd', 'btmanagerd')
 # Shared Wi-Fi/CONSYS combo processes — killing one can drop Wi-Fi. Never touch.
 _CRITICAL_COMMS = ('wmt_service', 'mtk_wmtd', 'wifid', 'wifim',
                    'stp_main', 'mtk_stp_psm', 'connfem')
-
-# Upstart rests in these; everything else (pre-start, spawned, stopping, ...) is
-# a transition, and evicting a job mid-transition leaves CONSYS half-configured.
-_SETTLED_STATES = ('running', 'waiting')
 
 
 def _find_bt_module(patterns=None):
@@ -115,66 +110,6 @@ def _find_holders(device_path):
     return holders
 
 
-def _job_state(job):
-    """(goal, state) of an upstart job, or (None, None) if it isn't known here."""
-    try:
-        r = subprocess.run(['/sbin/initctl', 'status', job],
-                           capture_output=True, timeout=5)
-    except Exception:
-        return (None, None)
-    if r.returncode != 0:
-        return (None, None)
-    head = r.stdout.decode(errors='replace').split('\n')[0].split(',')[0]
-    fields = head.split()
-    if not fields or '/' not in fields[-1]:
-        return (None, None)
-    goal, _, state = fields[-1].partition('/')
-    return (goal, state)
-
-
-def _wait_for_settle(jobs, timeout, poll=0.5):
-    """Block until every job leaves its transitional state, or timeout expires.
-
-    Starting on a boot event only says the CONSYS chip powered up, not that the
-    stock BT stack finished claiming it. Evicting acsbtfd mid-init flaps the
-    radio Wi-Fi shares, which is what boot-loops a Scribe (#87).
-    """
-    deadline = time.monotonic() + timeout
-    announced = False
-    while True:
-        unsettled = []
-        for job in jobs:
-            goal, state = _job_state(job)
-            if state is not None and state not in _SETTLED_STATES:
-                unsettled.append(f'{job} {goal}/{state}')
-        if not unsettled:
-            if announced:
-                log.info("Stock BT stack settled")
-            return True
-        if time.monotonic() >= deadline:
-            log.warning(f"Stock BT stack still unsettled after {timeout:.0f}s: "
-                        f"{', '.join(unsettled)}")
-            return False
-        if not announced:
-            log.info(f"Waiting for stock BT stack to settle: {', '.join(unsettled)}")
-            announced = True
-        time.sleep(poll)
-
-
-def _wait_for_stopped(job, timeout, poll=0.5):
-    """Block until an upstart job reaches stop/waiting, or timeout expires."""
-    deadline = time.monotonic() + timeout
-    while True:
-        goal, state = _job_state(job)
-        if state is None or (goal, state) == ('stop', 'waiting'):
-            return True
-        if time.monotonic() >= deadline:
-            log.warning(f"'{job}' did not reach stop/waiting in {timeout:.0f}s "
-                        f"(now {goal}/{state})")
-            return False
-        time.sleep(poll)
-
-
 def _graceful_kill(h, settle, device_path):
     """SIGTERM an unknown holder, then SIGKILL only if it clings to the node."""
     for sig in (signal.SIGTERM, signal.SIGKILL):
@@ -189,7 +124,7 @@ def _graceful_kill(h, settle, device_path):
         time.sleep(settle)
 
 
-def _evict_holders(device_path, settle, stop_timeout=10.0):
+def _evict_holders(device_path, settle):
     """Free device_path: stop stock BT via upstart, spare Wi-Fi, SIGTERM the rest."""
     holders = _find_holders(device_path)
     if not holders:
@@ -210,8 +145,6 @@ def _evict_holders(device_path, settle, stop_timeout=10.0):
         log.info(f"Stopping stock BT service '{job}' via upstart")
         if not run(['/sbin/initctl', 'stop', job]):
             log.warning(f"'initctl stop {job}' failed")
-    for job in stock_jobs:
-        _wait_for_stopped(job, stop_timeout)
     if stock_jobs:
         time.sleep(settle)
 
@@ -259,14 +192,8 @@ class MtkChip(BtChip):
             log.info(f"{device_path} is available")
             return True
 
-        log.info(f"{device_path} is busy, waiting for the stock BT stack to settle...")
-        _wait_for_settle(_STOCK_BT_JOBS, config.bt_stock_settle_timeout)
-        if _is_device_free(device_path):
-            log.info(f"{device_path} is available")
-            return True
-
         log.info(f"{device_path} is busy, identifying and evicting holders...")
-        _evict_holders(device_path, settle, config.bt_stop_timeout)
+        _evict_holders(device_path, settle)
         if _is_device_free(device_path):
             log.info(f"{device_path} is now available")
             return True
--- a/kindle_hid_passthrough/bt_mtk.py
+++ b/kindle_hid_passthrough/bt_mtk.py
@@ -6,7 +6,7 @@ import os
 import signal
 import time
 
-from bt_chip import BtChip, free_device, run
+from bt_chip import BtChip, run
 from config import config
 from logging_utils import log
 
@@ -14,6 +14,13 @@ DEFAULT_MODULE_PATTERNS = [
     'wmt_cdev_bt.ko',   # MediaTek (PW4/5, Kindle 10/11, Scribe)
     'bt_drv.ko',         # Older Freescale/NXP Kindles
 ]
+
+# Stock BT stack: respawn'd upstart jobs, so stop via upstart, never SIGKILL.
+_STOCK_BT_JOBS = ('acsbtfd', 'btmanagerd')
+
+# Shared Wi-Fi/CONSYS combo processes — killing one can drop Wi-Fi. Never touch.
+_CRITICAL_COMMS = ('wmt_service', 'mtk_wmtd', 'wifid', 'wifim',
+                   'stp_main', 'mtk_stp_psm', 'connfem')
 
 
 def _find_bt_module(patterns=None):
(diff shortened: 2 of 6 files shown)


mergen3107:
> if you can build this branch and test it on your Scribe that'd be a huge help

I have KS1 on 5.17.X
Is replacing the .upstart file with this one enough to test? I didn’t setup build environment yet

mergen3107:
Also, I disabled ALL native logging on KS1 since I despise Amazon snooping by disabling the four logger bins:
```

ln -s -b /dev/null /usr/sbin/syslog-ng
ln -s -b /dev/null /usr/sbin/loginfo
ln -s -b /dev/null /usr/sbin/tinyrot
ln -s -b /dev/null /usr/bin/showlog
```

Is this a problem for this project and/or testing?

zampierilucas:
@cosbot1 no need to build, if you've got the repo checked out just run `just install-ci fix/issue-87-mtk-boot-loop`, that pulls the compiled binary from the branch CI and installs it on the kindle for you. Would love to know if it clears the boot loop on your KS1, thanks for testing!

mergen3107:
Do you have a build with this fix? I am not sure I can do this (Windows)

zampierilucas:
@mergen3107 yes, the build is attached to this pr https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/28746031117 download kindle-hid-passthrough-armv7  and extract with something like 7z on windows then copy  it over, I know other managed to install from windows, so its doable, but yeah, should be better

mergen3107:
Ok, I copied the artifact, and installed options 4,5,6. Then restarted and now I am in bootloop...
Trying to catch it with SSH over SSH.

mergen3107:
OK, I was lucky to catch it 3-4 times out of like 30, to delete the file from the terminal.
I'll go and see if there are any useful logs left somewhere.

mergen3107:
Command I used was:
```
mntroot rw; rm -f /etc/upstart/hid-passthrough.conf; mntroot ro
```

mergen3107:
Update: it turned out that this command didn't go through in SSH.
I later deleted the conf file manually.
Don't really know why/how the boot loops stopped.

zampierilucas:
@mergen3107 can you pull any logs? I currently have no idea where in the stack the bootloop is :(

mergen3107:
Sure, I'll try. What's the best way to grab them? 

Like, do you have logs being written somewhere? Or is this Kindle logs?

zampierilucas:
@mergen3107 the one I care about is our own log, it's a plain file so your native-logger nuking doesn't touch it, and it survives reboots so the loop attempts should still be in there:

```
cat /var/log/hid_passthrough.log
```

that tells me if `prepare()` even ran and who was holding `/dev/stpbt`. Also grab the kernel log if you can:

```
dmesg | grep -iE 'consys|stp|wmt|wifi|wlan|mt76|bt' | tail -50
```

though dmesg wipes on each reboot so it's best-effort in a loop. Paste whatever you get. Thanks for grinding through this 😬

mergen3107:
Unfortunately, the `/var/log/hid_passthrough.log` log starts after boot loops. It seems that logs didn't even write during boot loops?
And the second log type is empty.


```
2026-07-15 16:51:13,325 INFO ble_hid: Preparing Bluetooth hardware...
2026-07-15 16:51:13,326 INFO ble_hid: Detected Kindle Scribe (code 0x875)
2026-07-15 16:51:13,328 INFO ble_hid: BT module already loaded: wmt_cdev_bt.ko
2026-07-15 16:51:13,378 INFO ble_hid: /dev/stpbt is available
2026-07-15 16:51:13,381 INFO ble_hid: API server listening on port 8321
2026-07-15 16:51:13,383 INFO ble_hid: Kindle HID Passthrough v3.8.0-ea69332 (daemon)
2026-07-15 16:51:13,387 INFO daemon: HID Daemon v3.8.0-ea69332
2026-07-15 16:51:13,388 INFO daemon: Device: FBX53C (DA:A2:04:98:ED:55) (ble)
2026-07-15 16:51:13,388 INFO daemon: === Starting connection ===
2026-07-15 16:51:13,391 INFO ble_hid: Devices: 0 Classic, 1 BLE
2026-07-15 16:51:13,391 INFO ble_hid: HID Host v3.8.0-ea69332
2026-07-15 16:51:13,391 INFO ble_hid: Opening transport...
2026-07-15 16:51:13,443 INFO ble_hid: Sending HCI Reset...
2026-07-15 16:51:13,630 INFO ble_hid: HCI Reset successful
2026-07-15 16:51:14,075 INFO ble_hid: Device powered on: 00:00:46:67:61:01/P
2026-07-15 16:51:14,086 INFO ble_hid: Controller address resolution enabled
2026-07-15 16:51:14,086 INFO ble_hid: BLE enabled
2026-07-15 16:51:14,088 INFO ble_hid: Keystore has 1 entries
2026-07-15 16:51:14,088 INFO ble_hid: Waiting for connection (Classic: 0, BLE: 1)
2026-07-15 16:51:14,106 INFO ble_hid: [BLE] Waiting for 1 device(s) (accept list)
2026-07-15 16:51:25,981 INFO ble_hid: [BLE] Device connected: FBX53C (DA:A2:04:98:ED:55/P)
2026-07-15 16:51:25,990 INFO ble_hid: [BLE] Restoring bonding...
2026-07-15 16:51:26,112 INFO ble_hid: [BLE] Bonding restored
2026-07-15 16:51:26,118 WARNING bumble.gatt_client: !!! received notification with no subscriber
2026-07-15 16:51:26,120 WARNING bumble.gatt_client: !!! received notification with no subscriber
2026-07-15 16:51:26,125 INFO device_cache: Loaded device cache for DA:A2:04:98:ED:55/P
2026-07-15 16:51:26,126 INFO ble_hid: Loaded cached descriptor (185 bytes)
2026-07-15 16:51:26,131 ERROR bumble.utils: [31m!!! Exception in wrapper:[0m
Traceback (most recent call last):
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/utils.py", line 336, in run
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/gatt_server.py", line 719, in on_att_find_by_type_value_request
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/gatt_server.py", line 719, in <genexpr>
RuntimeError: Task got bad yield: Attribute(handle=0x0001, type=UUID-16:2800 (Primary Service), permissions=1, value=0018)
2026-07-15 16:51:26,309 INFO ble_hid: [BLE] Found HID service
2026-07-15 16:51:26,714 INFO ble_hid: [BLE] Found input report 2
2026-07-15 16:51:26,893 INFO ble_hid: [BLE] Found input report 3
2026-07-15 16:51:26,983 INFO ble_hid: [BLE] Found input report 4
2026-07-15 16:51:27,073 INFO ble_hid: [BLE] Found input report 5
2026-07-15 16:51:27,077 INFO uhid_handler: Created UHID device: FBX53C (vendor=0x0000, product=0x0000, rd_size=185)
```

mergen3107:
I saw sometime ago an upstart watchdog sort of thing that counts unsuccessful bootloops and takes action to restore the boot. This might help debug it, and maybe can be added to the app to prevent any future trouble. Let me have a look

mergen3107:
Is there a specific reason that `start on started wmt` is used? Why not, for example, `start on started lab126_gui`? `lab126_gui` has a failed reboot counter (up to 3), so if it fails three times in a row it can make changes. Maybe this can be used to delete the upstart?

mergen3107:
Another idea is to add a delay:
```
start on started wmt

pre-start script
    sleep 30
end script
```

If one boot fails, the next one will give a chance to connect via SSH (hopefully). Downside is that each boot will be 30 seconds longer. Maybe only during testing should be enough.

zampierilucas:
@mergen3107 before I touch triggers I need to know if our job even runs during the loop, since your log shows `stpbt available` and no eviction. drop this as the first line inside the `script` block, right before `exec`:

```
echo "$(date) hid job started" >> /var/log/hid_passthrough.log
```

if it piles up during the loop, timing theory is live and `lab126_gui` is worth a shot. if it never shows, the loop's below us and none of these upstart tweaks matter. no rush 🙏

mergen3107:
@zampierilucas news!!
```
Sun Jul 19 00:20:27 +05 2026 hid job started
```
This is from the /var/log/hig_passtrough.log file, the rest of it is just normal daemon things, like waiting for device etc.

However, this is not the clean result :D let me explain.

> echo "$(date) hid job started" >> /var/log/hid_passthrough.log

I added this yes, but I also added my sleep delay above, so the whole script is now looking like this:

```
# HID Passthrough - Bluetooth HID host with UHID passthrough
# For use with pre-built binary from releases
start on started wmt
stop on stopping filesystems

pre-start script
    sleep 30
end script

respawn
respawn limit 5 60

env KINDLE_HID_BASE=/mnt/us/kindle_hid_passthrough

console none

script
    echo "$(date) hid job started" >> /var/log/hid_passthrough.log
    exec /mnt/us/kindle_hid_passthrough/kindle-hid-passthrough --daemon >>/var/log/hid_passthrough.log 2>&1
end script
```

However, I timed the boot and it started showing the progress bar only after 10-12 seconds. So the `sleep 30` did not occur? I am not sure now.

Do you think you recognize what's going on? Like, maybe Kindle needed just some more time for conflicting processes to resolve, or something.

Otherwise I can remove the sleep part and do the test again (if absolutely necessary, boot loops are pain).

mergen3107:
Full logs after start. Keyboard was recognized immediately, no need to open app even.
[hid_passthrough_NO_BOOT_LOOP.log.txt](https://github.com/user-attachments/files/30156157/hid_passthrough_NO_BOOT_LOOP.log.txt)

zampierilucas:
@mergen3107 that's the confirmation I needed, thanks. and your `sleep 30` did fire, the progress bar at 10-12s is the UI job coming up separately, the sleep only holds back our daemon. clean marker + no loop is a working delay.

so it's confirmed: starting on `wmt` grabs `/dev/stpbt` the instant the CONSYS chip powers up, flapping the radio wifi shares. `sleep 30` proves it but it's a band-aid. I swapped the trigger to your earlier idea, `start on started lab126_gui`, fires after wifi settles with no magic number, plus the 3-strike recovery for free. pushed to the branch.

build here once CI's green: https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/30208841014, grab `kindle-hid-passthrough-armv7`. **drop your `sleep 30` first** so we test the trigger, not the sleep. clean boot = I merge 🤞

mergen3107:
Thanks for having a look! I will have a look in 2-3 days if not sooner

mergen3107:
So I grabbed the newest build. Copied it to KS1, installed via scripts (this time I chose option 1, install everything). Double checked that the changes are there in the upstarts (start on gui and no 30 seconds delay). 

Now I got into a bootloop, but with new pattern:
- Progress bar starts
- It gets stuck at the beginning, after the seventh progress bar chunk.
- Kindle then refreshes the screen, progress bar is gone, but the three is still there, and wildly the warmth is cranked.
- Then it reboots.

I'll go and try restore from the bootloop...

mergen3107:
Ok, wasn't too bad. This time around there is ample time to connect via USB SSH (which is available at each loop) and execute my one-liner above:
```
mntroot rw; rm -f /etc/upstart/hid-passthrough.conf; mntroot ro
```

This time there was no log! `/var/log/hid_passthrough.log` is not present, so we don't know if it tried to start, maybe not?

There is however `kindle-button-mapper.log` that shows it started well:
```
[2026-07-26T17:57:20Z INFO  kindle_button_mapper] Kindle Button Mapper 1.3.0 (build 29955f1) starting...
[2026-07-26T17:57:20Z INFO  kindle_button_mapper] Config: devices=1, debounce=0ms, long_press=500ms, repeat=100ms
[2026-07-26T17:57:20Z INFO  kindle_button_mapper::vkeyboard] Virtual keyboard at /dev/input/event5 (target written to /var/run/kindle-button-mapper-key-target)
[2026-07-26T17:57:20Z WARN  kindle_button_mapper::layout] layout 'us,ru,kz' bound, but the running server did not reload it
[2026-07-26T17:57:21Z INFO  kindle_button_mapper::input] Waiting for device to appear...
```

mergen3107:
BT manager daemon didn't auto start (obviously), but works fine after entering the app. Buttom mapper is also fine, it continued the log after device was connected:
```
[2026-07-26T18:00:03Z INFO  kindle_button_mapper::input] Found device at /dev/input/event6
[2026-07-26T18:00:03Z INFO  kindle_button_mapper::input] Grabbed device exclusively
[2026-07-26T18:00:03Z INFO  kindle_button_mapper::input] Reading events from FBX53C (uniq="DA:A2:04:98:ED:56/P")
[2026-07-26T18:00:03Z INFO  kindle_button_mapper] [fbx53c] device connected
[2026-07-26T18:02:10Z INFO  kindle_button_mapper] keep-awake: re-armed screensaver idle timer
```

mergen3107:
So it seems like just switching to `started on lab126_gui` didn't help.
@zampierilucas thoughts?

zampierilucas:
@mergen3107 yeah, lab126_gui was the wrong lever, the framework draws that progress bar so we already fired by the time it stalls. reverted it, back on wmt.

also the missing log doesn't tell us anything, /var is a tmpfs and gets wiped every boot, so you were reading the log of the boot after you deleted the job.

what your sleep 30 was actually buying is time for acsbtfd to finish coming up. we were stopping it mid pre-start and taking the combo chip down with it. so prepare() now polls initctl until acsbtfd/btmanagerd settle before touching /dev/stpbt, and half the time the node is free by then anyway. bounded at 45s, no magic number.

new build here https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/30303135196, grab kindle-hid-passthrough-armv7. drop the sleep 30 first. fingers crossed 🤞

zampierilucas:
@mergen3107 #157 to disable autostart by default

zhaoyongheng-cmyk:
Hello,

I need help recovering my 2022 Kindle Scribe.

The device was jailbroken and running KOReader. I installed the ARMv7 version of kindle-hid-passthrough and selected:

“Install or update everything (recommended)”

The installation appeared to complete successfully. However, when I opened BT Manager and tried to scan for Bluetooth devices, I received this error:

[Errno 16] Device or resource busy: '/dev/stpbt'

After rebooting the Kindle, it became stuck in an infinite boot loop. It repeatedly displays the Kindle logo and restarts.

I have already tried holding the power button for 40 seconds, but the problem remains.

The current situation is:

- The Kindle never reaches the Home screen.
- KOReader and KTerm cannot be opened.
- My Mac does not detect the Kindle as a USB device or storage drive.
- I cannot copy a DO_FACTORY_RESTORE file to the Kindle.
- A repair technician inspected the motherboard and said there is no obvious UART header or reserved serial connector.
- I do not want to randomly solder to unknown test points or flash the device without a confirmed recovery method.

The installation directory used was:

/mnt/us/khp-release/

The installation command was:

sh -x /mnt/us/khp-release/scripts/install.sh

Could you please advise:

1. Is there a Scribe-specific way to disable the hid-passthrough daemon or its auto-start service while the device is stuck in a boot loop?
2. Is there a documented recovery or diagnostic mode for the 2022 Kindle Scribe?
3. Does the Scribe have known UART or other recovery test points?
4. Is there any way to recover the device without USB storage access?
5. Could the '/dev/stpbt' resource-busy error have caused this boot loop?

I understand that this may require advanced hardware recovery. I would appreciate any confirmed procedure or information specific to the 2022 Kindle Scribe.

Thank you.

mergen3107:
If you didn't setup usbnetlite before that, I have bad news for you

zampierilucas:
Hi @zhaoyongheng-cmyk Im really sorry about that, I thought we had patched at least the infinite boot loop by removing the auto-start service by default, Im not sure what else could be causing it now.

1 - The only way found so far is what @mergen3107 suggested which over usb ssh and quickly do:
`mntroot rw; rm -f /etc/upstart/hid-passthrough.conf; mntroot ro`

2 3 4 - Honestly I dont know much about this device in particular, and if you read the mobile reads threads about it, you will find @mergen3107 and pretty much the expert on it, so if he has not found another way, unfortunately I doubt it.

5 - Unfortunately yes, users reported at least two scribe models with this issue, for whatever reason it bootloops.

Again, Im really sorry about bricking your device.

mergen3107:
There is a slim chance Kindle comes back to senses after you plug it in and wait. I took me a lot of boot loops before it came back

cosbot1:
@zhaoyongheng-cmyk the way i was fixing it through usbnetlite:
one terminal pinging 192.168.15.244 to see if the ~3 second window comes
and ran this .exp script on mac so i don't have to snipe that small window:


```
#!/usr/bin/expect -f
set IP "192.168.15.244"
set PASS "kindle"
set timeout 3

puts "Removing HID passthrough + KOReader extension"

while {1} {
    spawn ssh -p 22 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$IP
    expect {
        -re "(?i)password:" { send "$PASS\r" }
        timeout { after 300; continue }
        eof     { after 300; continue }
    }
    expect {
        "kindle us" {
            puts "shell ready"
            send "mntroot rw\r"
            expect "kindle us"
            puts "removing hid passthrough"
            send "rm -rf /mnt/us/hidpas\r"
            expect "kindle us"
            puts "removing koreader hid extension"
            send "rm -rf /mnt/us/extensions/koreader/plugins/hid_passthrough*\r"
            expect "kindle us"
            send "rm -rf /mnt/us/extensions/koreader/plugins/kindle-hid*\r"
            expect "kindle us"
            puts "checking koreader config for autostart hooks"
            send "grep -r 'hid' /mnt/us/extensions/koreader/settings/ 2>/dev/null | head -5\r"
            expect "kindle us"
            send "sync\r"
            expect "kindle us"
            puts "rebooting"
            send "reboot\r"
            exit 0
        }
        timeout { after 300 }
        eof     { after 300 }
    }
    after 200
}

```

Now I don't think the koreader plugin causes it, maybe the autostart, it is working fine when i just installed the btmanager service, the app and the koreader plugin.

mergen3107:
@zampierilucas 
Sorry, I was away for some time. Now I can test again.
Is the message here still relevant? https://github.com/zampierilucas/kindle-hid-passthrough/pull/99#issuecomment-5096735835
I see you made a few more changes, so I don't know if this is a good build to test

zampierilucas:
@mergen3107 I dont even remember where we stopped, let me read back the above, and diff what changed, as I did one or two major reworks under this

zampierilucas:
@mergen3107 yes, that comment is still the plan. I only rebased the branch onto current main, the fix itself is unchanged. grab the fresh build from https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/31892008290.

One heads up, auto-start is now opt-in (#157), so after installing pick option 5 (or flip Start on boot in BT Manager), otherwise the boot path never runs. that also rewrites the job file, so your sleep 30 edit goes away on its own.

if it boots clean, paste /var/log/hid_passthrough.log.

mergen3107:
`BTManager v3.13.1-9d2f39f`

I did option 1 (install everything) + 5 (upstart).

Nope, it didn't like it. Bootloop...

mergen3107:
Ok, I caught it from about 20th attempt :D

mergen3107:
FWIW, Installed Button Mapper 1.4.5 from here: https://github.com/zampierilucas/kindle-button-mapper-rs/releases/tag/v1.4.5
Used `sh install.sh` from within the folder. Its upstart (`/etc/upstart/kindle-button-mapper.conf`) is working fine...
It's `/var/log/knidle-button-mapper.log` is sane:
```
[2026-08-20T21:42:26Z INFO  kindle_button_mapper] Kindle Button Mapper 1.4.5 (build f1a0938) starting...
[2026-08-20T21:42:26Z INFO  kindle_button_mapper] Config: devices=1, debounce=0ms, long_press=500ms, repeat=100ms
[2026-08-20T21:42:26Z INFO  kindle_button_mapper::vkeyboard] Virtual keyboard at /dev/input/event4 (target written to /var/run/kindle-button-mapper-key-target)
[2026-08-20T21:42:26Z INFO  kindle_button_mapper::vkeyboard] No page buttons found, page turns go to the virtual keyboard
[2026-08-20T21:42:26Z INFO  kindle_button_mapper::vkeyboard] Key injection FIFO at /var/run/kindle-button-mapper-key.fifo
[2026-08-20T21:42:27Z INFO  kindle_button_mapper::layout] layout 'us,ru,kz' active
[2026-08-20T21:42:28Z INFO  kindle_button_mapper::input] Waiting for device to appear...
```

zampierilucas:
@mergen3107 since your sleep 30 is the only thing that ever worked, the job now starts on `framework_ready`, which fires when the boot splash goes away. very late in the bringup, but if it's stable, sure. it also disables itself after 3 bad boots and saves `boot.log` plus `last_kmsg.txt` to `/mnt/us/kindle_hid_passthrough/`, so no more recovery sniping and we finally get evidence if it loops.

build here https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/32421936781, install option 1 then 5. if it loops, let it do its 3 attempts and send me those two files.

mergen3107:
Ok, I'll see if it can finish before I go to sleep

zampierilucas:
@mergen3107 yeah dont worry, and it sucks that this build takes 20 minutes every time, let me get a faster runner to maybe speed this up

mergen3107:
# OK, success, no more bootloops!

It seems like it booted fine after a restart, right away.

There are now three new files in the folder:
`boot.log`:
``
Fri Aug 21 03:30:37 +05 2026 start attempt 1
``

`boot_attempts`:
``
1
``


and `last_kmsg.txt` (30KB): [last_kmsg.txt](https://github.com/user-attachments/files/31283343/last_kmsg.txt)

Interestingly, `boot_attempts` file disappeared shortly after I opened it. Maybe BTdaemon started up and cleaned it up. The other two files are still present.

Anything else I can do for testing?

zampierilucas:
@mergen3107 I think thats it, thank you so much for all the help over months, I have pushed a cleanup commit and will get this merged and out there, so hopefully we dont brick any other kindle

mergen3107:
Cool! Thank you for all your patience and hard work. Thanks for bringing BT devices to Kindle!!

I'll probably test it one more time to make sure nothing extra was removed :D 

See you in the other (less stressful) tickets

mergen3107:
By the way, is it worth adding the same boot counter for Button Mapper's upstart?

I am afraid we never know if Amazon decided to change something and this would cause a bootloop. Seems like a good safety net in general

zampierilucas:
@mergen3107 good idea, same breaker for the mapper is up in zampierilucas/kindle-button-mapper-rs#54

mergen3107:
> I'll probably test it one more time to make sure nothing extra was removed :D

Tested version [edbacf3](https://github.com/zampierilucas/kindle-hid-passthrough/releases/tag/dev-edbacf3) and it works fine, no bootloops.

Issue #85, opened 2026-07-02
Oasis3: 2nd connection to 8bit zero2: no input dev generated.


danergo:
I'm running this on Oasis3 (10th gen I think).

There are some issues still for me:

"8BitDo Zero 2 gamepad??????????" is displayed in BT Manager. I have no idea why the name contains 10 "?" icons.

A bigger one: after first pairing, "/dev/lib/event10" is created. However after second connection, event dev is not being created.

zampierilucas:
Hey @danergo, I can't do much here without a log of it actually failing.

Fresh start, follow the log, then power-cycle the gamepad so it reconnects and paste from where it connects onward:

```bash
stop hid-passthrough; start hid-passthrough
tail -f /var/log/hid_passthrough.log
```

Also `cat /mnt/us/kindle_hid_passthrough/devices.conf` so I know if it's BLE or Classic. Without that I'm just guessing.

danergo:
Hi @zampierilucas:

Finally I'm with my computer to post you real logs from this Kindle.

`hid_passthrough.log` for the first connection:

> # cat /var/log/hid_passthrough.log 
> 2026-07-10 16:28:08,555 INFO ble_hid: loading uhid-4.1.15-lab126-443455-zelda.ko
> 2026-07-10 16:28:08,628 INFO ble_hid: Preparing Bluetooth hardware...
> 2026-07-10 16:28:08,629 INFO ble_hid: Detected Kindle Oasis 3 (code 0x3D7)
> 2026-07-10 16:28:08,662 INFO ble_hid: Warming BCM chip via btfd (BTenable)...
> 2026-07-10 16:28:09,759 INFO ble_hid: bsa_server up; BCM chip warmed
> 2026-07-10 16:28:11,790 INFO ble_hid: Froze btd (31874) so it won't respawn bsa_server
> 2026-07-10 16:28:11,817 INFO ble_hid: Killed bsa_server (31885); taking the UART
> 2026-07-10 16:28:12,974 INFO ble_hid: /dev/ttymxc2 ready (warm handoff complete; wake deferred to open)
> 2026-07-10 16:28:12,988 INFO ble_hid: API server listening on port 8321
> 2026-07-10 16:28:12,993 INFO ble_hid: Kindle HID Passthrough v3.7.0-cecdc3b (daemon)
> 2026-07-10 16:28:12,998 INFO daemon: HID Daemon v3.7.0-cecdc3b
[...]

PR #104, opened 2026-07-11
fix(classic): drop invalid bytes from SDP device name instead of replacing

zampierilucas:
Refs #85 (fixes the garbled name, the reconnect problem is separate and gets its own PR)

The 8BitDo Zero 2 pads its SDP service name attribute with junk bytes that aren't valid UTF-8. We were decoding with errors='replace', which faithfully turns every junk byte into U+FFFD, so the name showed up as "8BitDo Zero 2 gamepad??????????" in BT Manager, dmesg and the UHID device name.

Fix is two small changes in the decode:
1. truncate at the first NUL, since fixed length name buffers are usually NUL padded C strings
2. decode with errors='ignore' so any remaining invalid bytes get dropped instead of rendered as replacement chars

Valid names in any language pass through untouched, and if the result ends up empty we keep the previous fallback name.

Tested the decode against the reported name with trailing garbage bytes, NUL padding, and a non-ASCII name ("Clavier Français"), all come out clean.

### DIFF
--- a/kindle_hid_passthrough/classic.py
+++ b/kindle_hid_passthrough/classic.py
@@ -443,8 +443,11 @@ class ClassicMixin:
                             try:
                                 name = attr.value.value
                                 if isinstance(name, bytes):
-                                    name = name.decode('utf-8', errors='replace')
-                                self.device_name = str(name)
+                                    name = name.split(b'\x00')[0].decode(
+                                        'utf-8', errors='ignore')
+                                name = str(name).strip()
+                                if name:
+                                    self.device_name = name
                             except Exception:
                                 pass
 


ZongbingHou:
@zampierilucas Recently fixed a lot of issues, hope to merge and release an official version.

zampierilucas:
@ZongbingHou I was waiting for this to get tested, but oh well, seems safe enough, we can bisect if needed

danergo:
@zampierilucas this one seems need a little more fixing:

- garbaged name was saved into devices.conf
- even after manually removed invalid bytes from devices.conf, on the UI in connection details, device name still displays invalid characters (title is not, due to removal from devices.conf).

Thank you!

zampierilucas:
@danergo good catch, thanks for keeping on this. Turns out #104 only fixed half of it. The SDP decode I patched isn't actually the name that gets saved, the one written to devices.conf comes from the scan result, and the scanner (plus the BLE paths) were still decoding with `errors='replace'`, so the junk bytes came straight through. And the name in connection details is the UHID device name, which gets baked in at connect time from that same garbled value, which is why cleaning devices.conf by hand didn't help it.

#116 pulls all the name decoding into one `clean_device_name()` helper so no path can spit out those replacement chars again. One thing when you test: it stops new garbage but a name already sitting in the cache/on a connected device sticks around, so do a `just clear-cache` and reconnect after grabbing it. Fingers crossed this is the last of it, thanks for testing!

Issue #107, opened 2026-07-14
Launcher nitpicks


mergen3107:
Both apps (kindle-hid and button-mapper) have the same two nitpicks:
1) ~~Once launched, their icons don't move up to the top as recently launched. E.g., see how KUAL or KOReader jump up.~~ See https://github.com/zampierilucas/kindle-hid-passthrough/issues/107#issuecomment-5033675526
2) Once exited, they do not redraw the top bar fully. For example, the arrow for top menu settings isn't draw, but still can be pressed.

zampierilucas:
@mergen3107 thanks for flagging these! the second one was honestly my own fault. I was hand rolling a custom header and close button and letting the app take over the whole screen, so it was covering the status bar strip, when native chrome gives you the status bar, title and close for free. dumb. should've just used that from the start :P

fix is in #108, switches BTManager to native chrome. did the same for button-mapper since it had the exact same issue.

the recency one (icons not jumping up to recently accessed) I haven't cracked yet, still poking at it.

mergen3107:
Thanks for having a look at this!

> the recency one (icons not jumping up to recently accessed) I haven't cracked yet, still poking at it.

https://www.mobileread.com/forums/showthread.php?t=314700

See if this provides some hints.
I can also ask around in Kindle Modding Community discord

zampierilucas:
the status bar redraw (#2) is fixed in #108. for the recency one (#1) — retested on my basic 5 and it works: launching either app bumps it to the top of the recent sort (`p_lastAccess` in cc.db updates to launch time). can't repro icons not moving up. closing, but reopen if you still hit it and include your model + fw version.

mergen3107:
> launching either app bumps it to the top of the recent sort

Sorry about this! It seems I overlooked the fact that on KS1 with Sanctuary JB **none** of the scriplet jump up at all. Fix is imminent next week.

mergen3107:
> on KS1 with Sanctuary JB none of the scriplet jump up at all

Corrections! 
1) KS1 is not Sanctuary, but rather OG WinterBreak with hotfix 1.11.0
2) Installing hotfix 2.3.7 from Kindle Modding Hotfix GitHub page fixed this!

Sorry for the false alarm.

PR #108, opened 2026-07-14
Use native chrome so the BTManager status bar renders

zampierilucas:
Fixes the top bar nitpick from #107. Two symptoms, same root cause: while BTManager was open the status bar (clock, wifi, battery) only showed up when one of those values changed, and on exit the home top bar didn't redraw fully, the settings chevron was missing but still tappable.

The WAF had configureSearchBar=none, which lets the app window own the whole screen including the status bar strip. The app's opaque background sat on top of pillow's bar, so nothing showed until pillow repainted a widget on its own and punched through. Same story on exit with the custom quit path.

So I stopped fighting it and switched to configureSearchBar=system, like KindleForge does, and dropped BTManager's own header. Native chrome now owns the status bar, the title and the close button, and the app content starts below the reserved strip instead of covering it. I also ripped out the redraw workaround I'd tried (pillow poke + fbink flush) and the now dead quit endpoint, native close handles teardown and redraws home clean.

Tested on a Basic 4 (FW 5.16.2.1.1): full status bar during the app, single native close, home redraws on exit.

This only covers the second nitpick. The recency one (icons not jumping to recently accessed) is still open.

Refs #107

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,6 +6,17 @@ Userspace Bluetooth HID host for Kindle with UHID passthrough.
 
 The Kindle is accessed via SSH using the host alias `kindle`.
 
+## Keep Awake During Development
+
+Stop the Kindle from sleeping/screensaving while developing on it:
+
+```bash
+ssh kindle "lipc-set-prop -i com.lab126.powerd preventScreenSaver 1"   # keep awake
+ssh kindle "lipc-set-prop -i com.lab126.powerd preventScreenSaver 0"   # restore
+```
+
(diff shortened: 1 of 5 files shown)


mergen3107:
I don't think this has worked.
I uninstalled everything from previous build and put this one. Installed the BT Manager app and keyboard service again, keyboard works. However, the top bar is still missing. Both while during the app is open and after.

zampierilucas:
@mergen3107 I think I know what that is, there is a waf caching layer that is a pain to properly clean up, let me try to bake something in the install script to cleanthis up

PR #110, opened 2026-07-16
Reload BTManager chrome via graceful appmgrd stop so config changes take effect

zampierilucas:
Follow-up to #108. mergen3107 reported that after uninstalling the old build and installing the #108 build, the status bar was still missing — the chrome fix didn't take.

Root cause: the "WAF caching layer" isn't a disk cache, it's the resident mesquite process. BTManager is served straight from disk, but mesquite keeps the app process alive in the background and appmgrd just foregrounds it on relaunch, so it never re-reads `config.xml`. The old process keeps whatever chrome it parsed at startup — in this case the pre-`configureSearchBar=system` hidden status bar — even after the files are updated.

Fix: unload the app via appmgrd (`lipc-set-prop com.lab126.appmgrd stop app://<id>`) on every install path so the next launch spawns fresh and re-reads config:
- `illusion/BTManager.sh` — the scriptlet KindleForge relies on (bundled in the release tarball, runs `install_waf` on launch). This is the one that reaches KindleForge users; it never called `install-waf-app.sh`.
- `illusion/install-waf-app.sh` — manual/dev install.
- `scripts/install.sh` — the interactive installer's uninstall path.

Uses the **graceful appmgrd stop, not `pkill`**: killing mesquite out from under appmgrd makes it report a crash ("Application Error") to the user. The `stop` is the same path the native X close button uses — it fully unloads the process cleanly. The app persists no client state (`saveContext=no`, no `localStorage`), so the reload is safe on every run and also covers in-place upgrades (where the appreg entry stays and `install_waf` would otherwise skip). The per-app mesquite storage-dir purge is kept defensively (mesquite names dirs by app id).

## Testing

Verified on a Basic 5 (FW 5.16.2.1.1) with a controlled repro:

| step | state | result |
|------|-------|--------|
| old build | `config=none`, fresh process | heading at top, no native close button / reserved strip |
| new build, no unload | `config=system` on disk, same resident PID | still stale — identical to old build (the bug) |
| scriptlet w/ appmgrd stop | stop unloads → new PID re-reads config | full status bar + native X + reserved strip render, **no crash dialog** |

Confirmed `pkill` triggers the "Application Error" crash dialog whereas `appmgrd stop` unloads cleanly (returns to the previously-resident app with no error).

Refs #107 #108

### DIFF
--- a/illusion/BTManager.sh
+++ b/illusion/BTManager.sh
@@ -51,6 +51,28 @@ EOF
     fi
 }
 
+# ---- Purge stale WAF cache ----
+
+purge_waf_cache() {
+    # mesquite caches the parsed chrome config and rendered assets per-app under
+    # /var/local/mesquite. That cache survives reinstalls and in-place upgrades,
+    # so an old entry keeps the previous chrome (e.g. a hidden status bar) even
+    # after the app files change. The app persists no client state, so purge on
+    # every launch to guarantee config.xml changes take effect.
+    MESQUITE_DIR="/var/local/mesquite"
+    # Ask appmgrd to unload the app so the next launch spawns fresh and re-reads
+    # config.xml. Use the graceful stop, not pkill: killing mesquite out from
+    # under appmgrd makes it report a crash ("Application Error") to the user.
+    lipc-set-prop com.lab126.appmgrd stop "app://$APP_ID" 2>/dev/null
+    sleep 1
+    [ -d "$MESQUITE_DIR" ] || return
+    for d in "$MESQUITE_DIR"/*[Bb][Tt][Mm]anager* "$MESQUITE_DIR/$APP_NAME" "$MESQUITE_DIR/$APP_ID"; do
+        [ -e "$d" ] || continue
+        rm -rf "$d"
+        log_msg "Purged WAF cache: $d"
+    done
+}
+
 # ---- Start helper daemon ----
 
 start_helper() {
@@ -84,6 +106,9 @@ log_msg "BTManager scriptlet starting"
 
 # Install/update WAF files
 install_waf
+
+# Drop stale WAF cache so config.xml chrome changes render
+purge_waf_cache
 
 # Start the HTTP helper
 start_helper
--- a/illusion/install-waf-app.sh
+++ b/illusion/install-waf-app.sh
@@ -46,9 +46,30 @@ echo "2. Setting permissions..."
 chmod +x "$SCRIPTLET"
 echo "   Done"
 
+# ---- Purge stale WAF cache ----
+#
+# mesquite caches the parsed chrome config and rendered assets per-app under
+# /var/local/mesquite. That cache survives an uninstall/reinstall, so an old
+# entry keeps the previous chrome (e.g. a hidden status bar) even after the
+# app files are updated. Purge it on every install so config.xml changes take.
+echo "3. Purging stale WAF cache..."
+MESQUITE_DIR="/var/local/mesquite"
+# Unload any running instance via appmgrd so the next launch re-reads config.xml.
+# Not pkill: killing mesquite makes appmgrd report a crash to the user.
+lipc-set-prop com.lab126.appmgrd stop "app://$APP_ID" 2>/dev/null
+sleep 1
+if [ -d "$MESQUITE_DIR" ]; then
+    for d in "$MESQUITE_DIR"/*[Bb][Tt][Mm]anager* "$MESQUITE_DIR/BT Manager" "$MESQUITE_DIR/$APP_ID"; do
+        [ -e "$d" ] || continue
+        rm -rf "$d"
+        echo "   Removed $d"
+    done
+fi
+echo "   Done"
+
 # ---- Register in appreg.db ----
 
-echo "3. Registering app..."
+echo "4. Registering app..."
 if [ -f "$APPREG_DB" ]; then
     # Rewritten every time, an existing registration can still be pointing at a
     # stale command or app path after an update.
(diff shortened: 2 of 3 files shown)


mergen3107:
I tried this artifact. Copied all files and reinstalled the BT Manager manually (selected 6).
Top bar is still empty, nothing is drawn there.
Although the installation log seems fine:
```
[root@kindle kindle_hid_passthrough]# sh scripts/install.sh

Select an option:
 1) Install everything (recommended)
 2) Pair Bluetooth keyboard
 3) List paired devices
 4) Install udev rules (keyboard service)
 5) Install upstart (auto-start on boot)
 6) Install BTManager app
 7) Install KOReader plugin
 8) Uninstall everything
 9) Quit
Enter choice [1-9]: 6
 -> Installing BTManager app

=== BTManager Installer ===

1. Checking WAF app files...
   Files at /mnt/us/kindle_hid_passthrough/illusion/BTManager
2. Setting permissions...
   Done
3. Purging stale WAF cache...
   Removed /var/local/mesquite/com.lzampier.btmanager
   Done
4. Registering app...
   Already registered
5. Installing scriptlet...
   Installed to /mnt/us/documents/BTManager.sh
6. Starting daemon...
   Daemon started

=== Installation Complete ===

You can now:
  - Open 'BT Manager' from the Kindle library (scriptlet)
  - Or launch directly: lipc-set-prop com.lab126.appmgrd start app://com.lzampier.btmanager

To start on boot, install the upstart config:
  cp /mnt/us/kindle_hid_passthrough/illusion/../assets/hid-passthrough.upstart /etc/upstart/hid-passthrough.conf

```

zampierilucas:
Thanks for testing @mergen3107! That log is actually really useful, it tells me the reload path did its job: step 3 unloaded the app via appmgrd and wiped the mesquite cache (`Removed /var/local/mesquite/com.lzampier.btmanager`). So the stale-process theory this PR was chasing isn't what's biting you, the process definitely came up fresh and re-read config on the next launch.

Which means either the config.xml on your device doesn't actually have the chrome fix, or `configureSearchBar=system` renders differently on your firmware. Could you paste a couple things so I can tell which:

```sh
grep -A2 chrome /mnt/us/kindle_hid_passthrough/illusion/BTManager/config.xml
cat /etc/prettyversion.txt 2>/dev/null || cat /etc/version.txt
```

The first should show `<kindle:asset key="configureSearchBar" value="system" />`. If it doesn't, the copy grabbed an old config.xml and that's the whole problem. The second tells me your model/firmware, I verified this on a Basic 5 on 5.16.2.1.1 so if you're on something different that's a likely culprit.

Also just to confirm, after the reinstall did you fully reopen BT Manager from the library? The installer only restarts the daemon, the fresh config read only happens when you relaunch the app itself.

mergen3107:
First thing - yes, confirm:
```
[root@kindle us]# grep -A2 chrome /mnt/us/kindle_hid_passthrough/illusion/BTManager/confi
g.xml
    <kindle:chrome>
        <kindle:asset key="configureSearchBar" value="system" />
    </kindle:chrome>

    <kindle:gestures>
--
        <param name="chrome" value="yes" />
    </feature>

--
        <kindle:app name="com.lab126.chromebar" value="yes" />
        <kindle:app name="com.lab126.btfd" value="yes" />
    </kindle:messaging>
```

Second thing - yeah. I am on 5.17.2:
```
[root@kindle us]# cat /etc/prettyversion.txt 2>/dev/null || cat /etc/version.txt
Kindle 5.17.2 (~~otaVersion~~)
```

mergen3107:
> Also just to confirm, after the reinstall did you fully reopen BT Manager from the library?

Yes, before each copying/installation I exit to Kindle menu. The scriplet updates and jumps to front, then I launch it.

zampierilucas:
Hey @mergen3107, config's right and the reload works, so the stale process was never your problem. Turns out `configureSearchBar=system` only reserves the strip, the painting is done by `JunoStatusBarDriver` over `com.lab126.chromebar`. On my 5.16.2.1.1 the firmware fills it in with no config from our app at all, my guess is 5.17.2 dropped that fallback.

With BT Manager open, could you paste the output of these?

```sh
lipc-get-prop com.lab126.chromebar lastChromeConfig
lipc-get-prop com.lab126.winmgr chromeState
grep -iE 'chromebar|juno' /var/log/messages | tail -30
```

Then, still in the app, does this make the clock and battery show up?

```sh
lipc-set-prop com.lab126.chromebar replayChromeConfig 1
```

If it does I know what to bake into the scriptlet. Also is the top bar blank after exiting KUAL or KOReader too, or only BT Manager? And you're on a KS1 while I'm on a Basic 5, so this might be device specific rather than firmware.

mergen3107:
Tested on `BTManager v3.13.1-9d2f39f` from https://github.com/zampierilucas/kindle-hid-passthrough/pull/99#issuecomment-5361740147

> lipc-get-prop com.lab126.chromebar lastChromeConfig

Returns empty line

> lipc-get-prop com.lab126.winmgr chromeState

0

> grep -iE 'chromebar|juno' /var/log/messages | tail -30

Cannot test, because I nuked all the logs on Kindle :D 

>  Then, still in the app, does this make the clock and battery show up?
lipc-set-prop com.lab126.chromebar replayChromeConfig 1

No. FWIW, Clock and battery come up by themselves once time or battery state changes.

mergen3107:
> Also is the top bar blank after exiting KUAL or KOReader too, or only BT Manager?

Only BTManager. After exiting from KUAL and KOReader everything is restored as normal in Kindle UI

mergen3107:
The same on 3.14.0

Issue #100, opened 2026-07-06
The Bluetooth page-turning operation will time out and go to sleep


ZongbingHou:
When I use a Bluetooth page turner to operate Koreader, the device automatically goes into sleep mode, just like it does after a long period of inactivity. I'm only using the page turner right now, with the KindlebtController.koplugin. I'm not sure if using it with kindle-button-mapper-rs would also cause it to go into sleep mode.

ZongbingHou:
And there’s another issue: when connecting a Bluetooth page turner, if there’s no operation for about two minutes, pressing the button will have a brief delay of a few seconds before responding. I’m not sure if it’s a problem with kindle-hid-passthrough or if my Bluetooth page turner is just like that, but I think no Bluetooth page turner should be designed this way. Not operating for two minutes is normal, and there shouldn’t be such a long delay when you press it afterward.

ZongbingHou:
Based on my own testing, the problem seems to still be with kindle-hid-passthrough, and it's not that there's a noticeable delay if you press the button again about two minutes later, but rather that even if you wait just a few seconds, the next button press will still have some delay.

ZongbingHou:
```
[root@kindle root]# dmesg | grep -i "blue\|hci\|bt" | tail -100
[    0.287865] Bluetooth: Core ver 2.20
[    0.288005] Bluetooth: HCI device and connection manager initialized
[    0.288026] Bluetooth: HCI socket layer initialized
[    0.288043] Bluetooth: L2CAP socket layer initialized
[    0.288085] Bluetooth: SCO socket layer initialized
[    0.500761] bt_pwrctrl_init called..
bt_reset pin obtained is 74
bt_dev_wake pin obtained is 72
bt_host_wake pin obtained is 73
[    0.501194] IRQ# for BT HOST WAKE: 102
[    0.505301] bt_pwrctrl_probe probe successful
[    0.531006] ehci_hcd: USB 2.0 'Enhanced' Host Controller (EHCI) Driver
[    0.531059] ehci-mxc: Freescale On-Chip EHCI Host driver
[    1.124962] Bluetooth: HCI UART driver ver 2.3
[    1.125073] Bluetooth: HCI UART protocol H4 registered
[    1.125104] Bluetooth: HCI UART protocol BCSP registered
[    1.125115] Bluetooth: HCI UART protocol ATH3K registered
[    1.125849] usbcore: registered new interface driver btusb
[...]

PR #114, opened 2026-07-17
feat(uinput): load bundled uinput.ko for Oasis 3

zampierilucas:
Adds uinput support to the existing kmod-loading path so the Oasis 3 gets a working `/dev/uinput`. Comes out of zampierilucas/kindle-button-mapper-rs#7, where the Oasis 3 hits `uinput device create failed: No such device (os error 19)`.

The Oasis 3 (zelda) is the one Kindle whose stock kernel ships without `CONFIG_INPUT_UINPUT`. hid-passthrough already carries and loads the per-device kmods, so this reuses that machinery rather than making button-mapper carry modules of its own.

- `ensure_uinput()` mirrors `ensure_uhid()`: no-op when `/dev/uinput` already exists (every other Kindle, uinput built in); otherwise glob `uinput-{kernel}-*-{codename}.ko`, `insmod`, create the node from sysfs if there's no devtmpfs.
- Wired into `prepare_bt()` right after `ensure_uhid()`, so it loads at daemon start.
- `_log_missing_kmod()` gains a `mod` param so the "no bundled X.ko" error names the right module.
- Ships `uinput-4.1.15-lab126-443455-zelda.ko`, cross-compiled against the matching GPL kernel source. vermagic `4.1.15-lab126 SMP preempt mod_unload modversions ARMv7 p2v8`, matches the reporter's `uname -a`. Deploy/CI already glob `modules/*.ko`, so packaging needs no change.


zampierilucas:
@danergo moved it here, even tho this kmod is not necessarily needed for hid, it is to actually pass button presses, and since the mechanism to load kmods is already here :shrug:

ZongbingHou:
@zampierilucas  Can you merge this PR into the main branch? I'm not sure if this is what's causing issue #100.

PR #121, opened 2026-07-25
Mouse cursor implementation

kbarni:
The mtk (display) driver has no mouse cursor plane. When a mouse is connected through Kindle-HID-Passthrough, it works (emits events, you can click etc), but can't see where the pointer is.

This PR adds back the mouse pointer. It's a background process that listens to input events and draws a mouse cursor using fbink library.

I couldn't find a way to start it automatically when a mouse is connected, so I just added a toggle at the bottom - so the users can enable/disable the cursor manually when a mouse is connected.
I used the *bibata dark* cursor theme from Linux Mint - I find it more modern than the classic pointer.

Files added/modified:
- kindle-hid-passthrough/mousecursor.c - the source code for the daemon
- scripts/mousecursor - binary, built locally for armhf architecture (as the build system doesn't include the C compilers and libraries)
- illusion/ - some files modified to add the daemon to the interface.

The daemon works well on my PW5 / FW 5.18.5, but **probably worth more testing before adding to the final release**.
I don't know whether the library works on soft float devices, in that case an arm-el binary needs to be added too.

Here's the build job to test it directly: https://github.com/kbarni/kindle-hid-passthrough/actions/runs/30115945897


zampierilucas:
Hey @kbarni, this is awesome, the missing cursor has bugged me for a while. Nice work.

Two things before I can merge it:

1. The prebuilt `scripts/mousecursor` binary. I don't want a compiled blob committed, CI won't rebuild it and I can't verify what's in it. The build already cross compiles for armv7 in Docker, so let's compile `mousecursor.c` there and ship it in the release artifact instead. Happy to help wire that up if fbink's build is annoying.

2. The manual toggle. The daemon already knows when a device connects and what class it is, so this should just turn itself on automatically when a mouse shows up, no toggle. That's the behaviour I'd want.

Really cool addition, thanks for this :D

kbarni:
Glad you like it. There were a few messages on mobilereads on this issue, I thought better look into this. Took some time to investigate, glad I managed to solve it.

About the points you raised:
1. I agree the binary shouldn't go on the repo. Only your release builds automatically and the build system has no X11 development library included. I found it easier to build it using my local toolchain than to modify the github action.
The c file has the build instructions at the beginning. It only needs GCC compiler and X11 library.

2. I didn't find where the library identifies the connected device and its class. It will be cool to start this automatically - all you need is to link the binary to the *connect*/*disconnect* events for a mouse device instead of the toggle.
That said, I use HID devices occasionally. I always start the daemon from the booklet when needed, and one more click didn't feel complicated.

RN I don't think I'll have the occasion to work on this - we are more into leaving Bordeaux for a few days (if you follow the news - but don't worry, everyone's safe).

mergen3107:
> we are more into leaving Bordeaux for a few days

@kbarni stay safe!!!

zampierilucas:
@kbarni take care of yourself and your family first, hope you all stay safe. No rush on any of this.

I'll take the PR from here, wiring the X11 build into CI and hooking the cursor to the mouse connect/disconnect events so it's automatic. You've done the hard part. Cheers!

zampierilucas:
@kbarni hey, good news, got it working on two kindles here, a Basic 3 and a Basic 4. Hope things are calmer over there.

So I took your branch and sorted the two things I mentioned. The binary builds in CI now, no blob in the repo. The trick was building it against an old glibc, anything newer than 2.34 just won't load on the kindle, and then patching the ELF interpreter to `/lib/ld-linux.so.3`. Ends up as one self contained binary that uses the kindle's own libX11, works on both my devices. And it starts and stops by itself now, I parse the HID report descriptor on connect and if it's a mouse the overlay comes up, dies on disconnect. No more toggle.

Couple of fun ones while testing. The fbink path was hardcoded to `/var/local/kmc/armhf/bin/fbink` which doesn't exist on either of my kindles, so it was drawing the cursor and never refreshing the panel, I saw absolutely nothing and spent a while poking at it. It looks fbink up at runtime now. Also added an erase on exit, otherwise the cursor just sits there burned on the screen after you disconnect, and a GC16 pass after a few seconds idle to clean up the A2 ghosting.

It's all in #126 if you want to take a look. Runs really nice on the Basic 3, the Basic 4 is still a bit laggy, my guess is the mtk eink just can't keep up with the refreshes, messing with waveforms to see if I can get it smoother. Thanks again for building this, it's a really cool addition :D

PR #122, opened 2026-07-26
simplified installation

kbarni:
**Previous PR resubmitted.** Needed to update my fork and it diverged slightly from this commit so I had to drop my commit and the PR.

So this PR modifies the workflow file to build the release so it can be copied directly to the Kindle. No need to run `install.sh` to copy the files to the correct location. Still need to run it to copy the keyboard rules and the dev-is-keyboard to the correct location on the system partition.

Before applying this PR make sure the *installation script in KindleForge is updated*.


zampierilucas:
Hey @kbarni thanks you for the pr, closing as we moved to kpm packaging

Issue #106, opened 2026-07-13
Kindle Oasis 2 (5.16.2.1.1): HCI Reset always times out after successful Broadcom warm handoff


syedharris:
Environment
Device: Kindle Oasis 2 (9th Gen, codename cognac)
Firmware: 5.16.2.1.1
Kernel: 4.1.15-lab126
Jailbreak: WinterBreak
Project Version: kindle-hid-passthrough v3.8.0
Summary

I'm unable to pair any Bluetooth keyboard on a Kindle Oasis 2.

The Broadcom initialization sequence appears to complete successfully until the first HCI command is sent. The controller never acknowledges the initial HCI Reset, causing the initialization to fail with a timeout.

From my testing, this appears to be specific to the Oasis 2. I understand the project works correctly on the Oasis 3.

Steps to Reproduce
Install kindle-hid-passthrough
Launch BT Manager
Select Pair Bluetooth Keyboard
Wait for Bluetooth initialization
Expected Behavior

After the Broadcom firmware is loaded and the UART handoff completes, the controller should acknowledge the HCI Reset command and continue into pairing mode.

Actual Behavior

Initialization consistently fails immediately after the transport is opened.

The last successful log messages are:

Detected Kindle Oasis 2

Warming BCM chip via btfd (BTenable)...

bsa_server up; BCM chip warmed

Froze btd

Killed bsa_server

/dev/ttymxc2 ready (warm handoff complete)

Opening transport...

BCM chip woken (dev_wake asserted)

BCM bluesleep disabled (chip stays awake)

Sending HCI Reset...

The controller never responds.

After approximately 10–12 seconds:

HCI Reset timed out

TimeoutError
Bluetooth Initialization Sequence

From both the runtime logs and reviewing the source code, the Broadcom initialization appears to follow this sequence:

BTenable

↓

wait for bsa_server

↓

Broadcom firmware download

↓

firmware settle delay

↓

freeze btd

↓

kill bsa_server

↓

free UART

↓

open transport

↓

assert dev_wake

↓

disable bluesleep

↓

settle delay

↓

device.host.reset()

Every step before device.host.reset() appears to complete successfully.

The failure occurs on the first HCI command.

Investigation Performed

I inspected the relevant source files:

transport.py
bt_setup.py
bt_brcm.py
Transport

The transport opens successfully.

transport = await open_transport(spec)

chip().on_transport_open()

device = Device.with_hci(...)

await device.host.reset()

[...]

PR #124, opened 2026-07-27
fix(brcm): survive UART byte loss on deep-idle wake (#120)

### LINKED ISSUE #120: Oasis 2: cpuidle state2 causes intermittent UART/HCI corruption after BLE idle; disabling state2 fixes it
## Summary

On a Kindle Oasis 2, BLE HID reports become intermittently unresponsive at normal reading intervals (roughly 10–30 seconds between presses), while rapid presses usually work. The failure correlates with corruption on the BCM ↔ i.MX UART HCI transport (`/dev/ttymxc2`).

A controlled A/B/A/C test indicates that disabling **only cpuidle state2** on both CPUs, while leaving state0 and state1 enabled, prevents the observed corruption. `IGNBRK` alone improved recovery but did not eliminate the problem.

This appears related to, but more specific than, #75 and #106.

## Environment

- Device: Kindle Oasis 2 / 9th Gen (`cognac`, device code `0x34A`)
- Firmware: 5.16.2.1.1 (`002-juno_16020101_cognac_zelda-409745`)
- Kernel: `4.1.15-lab126`, armv7l
- SoC: i.MX7 Dual
- Bluetooth: Broadcom over `/dev/ttymxc2`
- UART speed reported by `stty`: 2,000,000 baud
- HID build: `v3.8.0-876208f` (PR #97 experimental build with UART parser resynchronization and 5-second HCI keepalive)
- Client: a paired BLE page-turn remote, used through KOReader

Device addresses and remote names are omitted from this report.

## User-visible symptom

- Repeated presses less than about 3 seconds apart are generally reliable.
- With normal reading intervals (often 10–30 seconds), a press may produce no page turn.
- The service may recover after several seconds or minutes without a manual restart.
- The Kindle and KOReader remain responsive during the failure.

## HCI evidence

During a normal-idle failure, the parser first saw an implausible length, then invalid packet types and zero-filled data. The 5-second keepalive timed out and recovered about 5 seconds later:

```text
01:41:27.391 implausible HCI body length 8256; resetting UART parser
01:41:27.410 invalid HCI packet type 0x14; resyncing
01:41:27.414 invalid HCI packet type 0x00; resyncing
...
01:41:31.968 HCI keepalive failed (1 consecutive): TimeoutError
01:41:32.781 Ignoring event [0x00]: 120000000000000000000000000000040e07...
...
01:41:36.xxx HCI keepalive recovered
```

The experimental resynchronization/keepalive logic is useful: it prevents a permanent hang and explains the automatic recovery. It cannot recover the HID report that was already corrupted/lost.

## Controlled test

For every phase:

- The same HID build, paired remote, configuration and reading workflow were used.
- `IGNBRK`, `-BRKINT`, `-PARMRK`, and `CRTSCTS` were re-applied one second after every `Opening transport...` event and read back for verification.
- `/proc/tty/driver/IMX-uart` was sampled every 2 seconds.
- Button Mapper events and HID logs were captured automatically.
- Original cpuidle values were restored after each phase.

| Phase | CPU idle condition | Duration | Successful button events logged | Longest gap between logged events | Mid-session UART/parser result |
|---|---|---:|---:|---:|---|
| A1 | normal idle | 18m 53s | 59 | 66s | no corruption observed in this run |
| B | state1 + state2 disabled | 25m 27s | 94 | 54s | clean |
| A2 | normal idle | 25m 40s | 69 | 48s | **22 invalid packet types, 1 implausible length, keepalive timeout + recovery** |
| C1 | **state2 only disabled** | 30m 30s | 86 | 127s | no parser or keepalive errors; one reason=8 disconnect occurred near the stop transition |
| C2 | **state2 only disabled** | 44m 10s | 130 | 241s | **clean** |

In phase C, the settings were:

```text
cpu0/state0/disable=0
cpu0/state1/disable=0
cpu0/state2/disable=1
cpu1/state0/disable=0
cpu1/state1/disable=0
cpu1/state2/disable=1
```

Normal idle reproduced missed input and HCI corruption. Both state2-only runs were subjectively reliable for every press, including long idle gaps, and the final ~44-minute run had no mid-session parser or keepalive errors.

Battery readings were too coarse and the sessions too short to make a reliable power comparison. State2-only should nevertheless be less intrusive than disabling both state1 and state2.

## Interpretation

Because all phases used the same `IGNBRK`/`CRTSCTS` handling and only the cpuidle condition changed, the result points to the i.MX side entering cpuidle state2 and failing to receive the beginning of the UART stream cleanly after wake. The zero-filled data is consistent with a break/wake transient or missed/corrupted RX bytes.

The IMX-uart `/proc` output on this kernel did not expose `brk`, `oe`, or `fe` counters, so the exact low-level failure mode could not be distinguished. It showed RX byte jumps at the failure time, but no explicit error counters.

This does not prove whether the ultimate cause is UART clock gating, wake latency, DMA/FIFO handling, or Broadcom firmware low-power behavior. However, the state2-only result strongly suggests a host power-management constraint can provide a practical Oasis 2 fix.

## Working workaround

A local lifecycle wrapper currently:

1. Saves the original `cpu*/cpuidle/state2/disable` values.
2. Sets only state2 `disable=1` while the HID UART transport is active.
3. Re-applies `IGNBRK -BRKINT -PARMRK CRTSCTS` after every UART reopen.
4. Restores state2 on explicit HID stop and during system suspend.
5. Disables state2 again after HID resumes.

This has made normal page-turn usage reliable so far without disabling state1.

## Possible upstream directions

Would you consider an Oasis 2-specific transport/power-management quirk while the BCM UART is active?

Possible approaches, in preferred order:

1. Hold an appropriate PM QoS / latency constraint while the HCI UART is open, if available on this old kernel.
2. Otherwise, save and disable cpuidle state2 for both CPUs while the Oasis 2 BCM transport is active, restoring the exact previous values on stop/suspend and reapplying on resume.
3. Ensure `IGNBRK` and `CRTSCTS` are asserted after every pySerial transport reopen.

I can provide a sanitized diagnostic bundle, test a debug build, or repeat longer state2-only/normal-idle comparisons if useful.

zampierilucas:
Fixes #120 #100

@yaqih0205-beep did the hard part here: a controlled A/B/A/C with the same build, remote, config and termios in every phase, varying only cpuidle. That isolates the trigger far better than anything I could have gotten remotely, and it's what this PR is built on.

## What's going on

Two independent problems stack up, which is why the symptom looked so strange (fast presses fine, a press after 10-30s of reading dropped, recovery minutes later without a restart).

**The SoC misses the start of the packet.** i.MX7D cpuidle `LOW-POWER-IDLE` powers off the ARM platform, switches to the 24 MHz OSC and puts DDR in self-refresh. Its exit latency is around 500 us. The UART RX FIFO is 32 bytes, which at 2 Mbaud is 160 us of slack, so the leading bytes are gone before the UART is clocked again. That also matches the reporter's observation that `/proc/tty/driver/IMX-uart` showed no `fe`/`oe`/`brk` counters: `uart_line_info()` only prints those when nonzero, so zero counters means the bytes were never sampled at all rather than sampled badly.

We make it worse ourselves. `_assert_wake()` writes `proto=0` to keep the chip awake, and Broadcom's `bluesleep_stop()` tears down the HOST_WAKE IRQ along with it, so nothing pulls the SoC out of idle ahead of the chip's data.

**Then the parser turns a lost byte into a lost session.** bumble trusts the length field, so the reported `implausible HCI body length 8256` silently swallows the next 8 KB of perfectly good HCI. At BLE HID traffic rates that is minutes of dead input, which is exactly the "recovers after several seconds or minutes" behaviour. An unknown packet type is no better: it raises out of `feed_data`, discarding the rest of that read.

## What this does

**Holds a CPU latency ceiling while the transport is open.** PM QoS `CPU_DMA_LATENCY` rather than the `state2/disable` sysfs flag, for three reasons: the constraint lives with the fd so it releases if the daemon is killed (the sysfs flag isn't refcounted and would leave state2 disabled forever, silently costing battery), it survives suspend/resume with no save/restore, and it doesn't assume state2 means the same thing on every model. Falls back to the sysfs flag with restore where the misc device is missing.

The budget is derived, not hardcoded: `32 bytes x 10 bits / baud x 0.5` = 80 us at 2 Mbaud. Against the i.MX7D table that allows `WFI` (1 us) and `WAIT` (50 us) and blocks `LOW-POWER-IDLE` (500 us) — the same state2-only outcome that was measured, reached from the hardware numbers. Where every state already fits the budget it takes no hold at all.

**Replaces the RX parser** with one that bounds every length field and resyncs a byte at a time instead of trusting a bad header. It accepts only ACL and Event: this is a HID host, it never sets up SCO or ISO, and excluding them denies noise two ways to look like a valid header. (Keeping SCO measurably hurt: `01 03 0c 00` desyncs into `03 0c 00 04`, a plausible-looking SCO header that then eats the following event.)

**Sets `IGNBRK` on the UART.** pyserial clears it on every open (`serialposix.py:423`), so a line break arrives as a literal `0x00` in the HCI stream — the reported `invalid HCI packet type 0x00`. This is why re-applying it after each reopen helped.

## Scope

Keyed off the chip backend rather than the transport spec string, because a `config.ini` override or the probe fallback hands the same UART back as a `file:` spec. MediaTek Kindles are untouched.

The parser and `IGNBRK` changes apply to every Broadcom Kindle and are pure robustness — correct regardless of whether the cpuidle theory holds. The latency hold is effective on i.MX7D (zelda: KOA2, KOA3) and self-disables on i.MX6SLL (rex: PW4, Basic 3) if its cpuidle driver only exposes WFI + WAIT. Since KOA3 is also zelda, this may touch the tail of the #75 / #106 flakiness too.

## Testing

14 local unit tests, no Kindle required (`just test`). They cover the 8256-byte swallow, resync after garbage, split reads, a stale partial packet, the i.MX7D state table, the sysfs restore path, and the never-force-WFI-only clamp. Two bugs got caught writing them: `log.exception()` doesn't exist on `HIDLogger` (would have thrown inside the exception handler), and the SCO desync above.

Not yet validated on hardware. I don't have a Broadcom Kindle that reproduces this — the cpuidle half needs an i.MX7D device, and my Basic 3 is i.MX6SLL where it should correctly do nothing. The parser swap point is verified by reading bumble (`common.py:296-298`, sink attached by `Device.with_hci` after we swap) but not executed; if that ordering were ever wrong it fails loudly as an HCI Reset timeout rather than silently.

The daemon logs its decision at startup either way:

```
grep -E "cpuidle budget|CPU latency|HCI resync" /var/log/hid_passthrough.log
```


ZongbingHou:
After merging #124 and running on Kindle Oasis 3 (i.MX7D), the lower-level UART/HCI corruption issues were resolved. However, when pressing the BLE remote button after a short idle interval (10–30 seconds), a noticeable delay of **1 to 3 seconds** was still observed before the page turned.

### Root Cause Analysis
During re-setup in `_setup_ble_hid()`:
1. GATT Service Discovery & Descriptor parsing were being executed repeatedly even when cached descriptors/reports were available, adding unnecessary GATT round-trip latency.
2. The `/dev/uhid` device was being recreated on repeated setup calls.
3. `await self._ble_activate_hid_service()` was synchronously blocking the setup pipeline. Writing to `GATT_HID_CONTROL_POINT` (*Exit Suspend*) and reading/writing `GATT_PROTOCOL_MODE` wait for GATT responses from the remote. When the remote is just waking up, these ATT exchanges often take 500ms–1.5s to respond, directly delaying the readiness of the HID input channel.

### Solution
This PR optimizes `_setup_ble_hid()` with the following improvements:
1. **Cache-first descriptor loading**: Prefers loading cached descriptors and existing `hid_reports` before triggering full GATT discovery.
2. **Idempotent UHID creation**: Skips recreating `uhid_device` if it already exists on the host instance.
3. **Non-blocking HID activation**: Dispatches `_ble_activate_hid_service()` as a background task (`asyncio.create_task`) and retains a reference on `self._activation_task` to prevent garbage collection in Python 3.8+.

### Verification
Tested on Kindle Oasis 3 (Firmware 5.18.2, i.MX7D) with a paired BLE page-turn remote:
- **Before**: 1–3s latency on the first keypress after ~10-30s of reading idle.
- **After**: Instant page-turn response on almost every idle keypress. The HID channel becomes active immediately after subscription without waiting for GATT Activation responses.

---

```python
async def _setup_ble_hid(self):
    """Discover reports, create UHID, subscribe. Common to connect and post-pair."""
    if not self.report_map:
        self._load_cached_descriptor()

    if not self.hid_reports:
        await self._discover_ble_hid_service(process_reports=True)

    if not self.report_map:
        raise InvalidStateError("[BLE] No report descriptor available")

    if getattr(self, 'uhid_device', None) is None:
        self._create_uhid_device()

    await self._subscribe_to_ble_reports()
    self._activation_task = asyncio.create_task(self._ble_activate_hid_service())

PR #126, opened 2026-07-28
feat(cursor): mouse pointer overlay for Kindle (builds on #121)

zampierilucas:
Builds on @kbarni's #121 and gets the mouse cursor overlay working end to end and CI-buildable. If you use a BT mouse with your Kindle, please grab the build and report how it feels.

## What it does
Kindles have no hardware cursor plane, so a connected mouse moves an invisible pointer. This paints a visible cursor: it polls the X pointer and draws to the framebuffer, refreshing via fbink. On top of kbarni's original:

- **auto on/off**: detects a mouse HID descriptor and starts/stops the overlay automatically on connect/disconnect (no manual toggle)
- **single CI-built binary**: compiled against old glibc with a patched interpreter, runs across Kindle models using the Kindle's own libX11. No committed blob, no bundle.
- **finds fbink at runtime** (KOReader / KMC / libkh), no hardcoded path
- **cleans up after itself**: erases the cursor on disconnect instead of leaving it frozen, and reaps orphaned overlays from restarts
- **de-ghosts**: a GC16 pass clears eink ghosting once you rest

## Testing
Grab the `kindle-hid-passthrough-armv7` artifact from the latest CI run on this PR, deploy it, pair a BT mouse, and report back with your model.

Tested on Basic 3 (rex) and Basic 4 (MTK): smooth on rex; the Basic 4 (Amazon's slowest) still gets a bit laggy from eink-update contention, tuning the waveform. Curious how PW5 and others do.


kbarni:
I tested the build from https://github.com/zampierilucas/kindle-hid-passthrough/actions/runs/30482660387
The binary won't run on my PW5 (like it was for another architecture). When replaced with my local build, it works.
The integration works well - bluetooth mouse is recognized and the daemon starts. The cursor is quite responsive, not perfect but usable..

zampierilucas:
@kbarni I have fixed the binary build on the latest push

kbarni:
@zampierilucas The binary works, so the build system is OK.

But the mouse **leaves a trail** on my device. I suspect it's because you switched to GC16 rendering from A2 (I don't see B/W artifacts), so redraw takes more time, but I prefer the cursor without trail.

(I'm leaving for holidays, so I won't be able to test during August)

Issue #101, opened 2026-07-06
Proposing a series of PRs: simultaneous BLE + Classic, and phone-as-keyboard


kal-kaliper:
Hi Lucas,

I've been running kindle-hid-passthrough on my Kindle and have built up a set of changes I'd like to contribute upstream. Before opening a pile of PRs, I wanted to lay out what's there and ask how you'd prefer to receive it.

Two headline capabilities:

- **Simultaneous BLE + Classic:** hold a BLE and a Classic HID session at the same time, so a phone over BLE and a physical keyboard over Classic can both drive the Kindle at once. This replaces the single-active-connection model with per-protocol sessions.
- **Phone-as-keyboard:** treat phone-class devices as HID candidates so Android keyboard / voice-to-text apps work as Bluetooth keyboards, with a small KOReader-plugin side that accepts them.

Everything else is hardening that fell out of making those two reliable:

- Classic reconnect isolation and retry/backoff
- BLE setup timeouts and recovery while Classic is unstable
- BLE/Classic radio coexistence (page-scan yielding, distinguishing Classic setup from a live session)
- Power-event (suspend/resume) lifecycle
- Runtime recovery after HID reconnects, and gating raw HID report bytes behind a diagnostics flag
- BLE "Kindle text mode" keyboard-report serialization

Plus two self-contained fixes I pulled out on their own:

- WAF debug-screen log sanitization (strip ANSI/control chars, render via textContent, so raw log bytes can't corrupt the view)
- Actively driving Classic `authenticate()` / `encrypt()` instead of passively waiting for the device to start

It has all been tested on real hardware (a phone and a keyboard connected at the same time), and each change is a focused commit with its rationale. I've also taken your earlier note about lean comments into account.

That's a lot, so rather than drop it all at once I'd like to send it as a series of small, focused PRs: probably starting with the self-contained ones (the log-sanitization fix is a good opener), then the simultaneous-sessions work as the foundation, then the pieces built on top, rebasing each onto main as the previous merges.

Does that approach work for you?

Thanks for the project,
Kal

zampierilucas:
Hey @kal-kaliper, thanks for laying it out first instead of dropping 12 PRs on me.

Send the self-contained ones, log sanitization, the active `authenticate()`/`encrypt()`, diagnostics gating. Those I'll review fast.

[...]

PR #131, opened 2026-07-30
fix(cache): key the device cache on the normalized address

kal-kaliper:
Found while working on the phone-as-keyboard branches in #101, but independent of them, so sending it on its own.

One device ends up with two cache files, and the wrong one is authoritative.

`_get_cache_path` keys on the raw address string. `str(connection.peer_address)` carries a transport suffix (`AA:BB:CC:DD:EE:FF/P`), and the Classic connection listener stores that in `current_device_address` for outbound connections as well as inbound — so `_load_cached_descriptor()` and `_query_classic_sdp()`, which both default to that field, read and write `<addr>_P.json`. The pairing flow and the startup descriptor check pass the plain `devices.conf` address instead, so they use `<addr>.json`.

What that costs today:

- **Removing a device leaves a stale descriptor behind.** `controller.py` normalizes the address before `clear()`, so only the plain file is deleted. The suffixed one survives, and after the device is paired again the connection path reads that stale copy in preference to what the new pairing just cached.
- **The descriptor cached at pairing is never read by the connection path.** On a first pairing that costs one redundant SDP query on the next connection.
- Two files per device, free to drift apart.
- The `Cached descriptor for ...` line at startup reports on the plain file, which the connection path does not use, so it can vouch for a descriptor that is not the one in play.

The fix normalizes with the existing `normalize_addr` before building the filename. `.replace('/', '_')` goes away with it, since there is no longer a suffix to escape.

That alone would orphan the pre-normalization files existing installs have, so there is a startup migration. Legacy names may carry a suffix, be lower case, or both; all three are folded into the canonical name. Where only the legacy file exists it is renamed, so a paired device keeps its descriptor instead of re-querying SDP. Where both exist as separate files the newer mtime wins.

One subtlety worth flagging for review: on a case-insensitive filesystem the legacy and canonical names can be the *same file*, and an earlier draft of this deleted the only copy in that case. There is now a `samefile` guard, with the reasoning in a comment.

I have no unit-test suite to add this to on your side, so verification was manual and is listed in the commit message. Happy to add tests wherever you'd want them.

### DIFF
--- a/kindle_hid_passthrough/device_cache.py
+++ b/kindle_hid_passthrough/device_cache.py
@@ -4,9 +4,21 @@
 import json
 import logging
 import os
+import re
 from typing import Dict, Optional
 
+from config import normalize_addr
+
 logger = logging.getLogger(__name__)
+
+# Pre-normalization filenames came straight from the address, so they may carry a
+# transport suffix (AA_BB_CC_DD_EE_FF_P.json for "AA:BB:CC:DD:EE:FF/P") and may be
+# in either case, because the old key was built without uppercasing. Both forms
+# have to be recognized, not just the suffixed one. The extension stays literal
+# lowercase: the old code only ever wrote ".json", so a ".JSON" file is somebody
+# else's, not a legacy cache entry.
+_LEGACY_KEY_RE = re.compile(
+    r'^([0-9A-Fa-f]{2}(?:_[0-9A-Fa-f]{2}){5})(?:_[PRpr])?\.json$')
 
 
 class DeviceCache:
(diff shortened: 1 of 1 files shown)


zampierilucas:
Hey @kal-kaliper, diagnosis is spot on and normalizing in `_get_cache_path` is the right fix, but can you cut it down to just that. The migration is more machinery than I want, and dropping it only costs an orphaned `<addr>_P.json` per paired device plus one redundant SDP query on the next connect.

fwiw the samefile guard wouldn't have held anyway, `/mnt/us` is fuse.fsp over vfat and reports different inodes for two casings of the same file, so it falls through to the mtime compare and deletes the only copy.

Also trim the comment blocks, that reasoning is already in the commit message and thats where I'd rather keep it. Should land around 3 lines.

zampierilucas:
landed in #174, your fix as written. thanks for the diagnosis, the two-files-per-device write-up is what made it obvious.

PR #132, opened 2026-07-30
fix(deploy): make deploy work from macOS, and actually stop the WAF app

kal-kaliper:
Three faults in the deploy path, found while deploying from a Mac. Independent of the #101 work.

**1. `deploy` cannot run from macOS at all.** The copy step rewrites paths with GNU tar's `--transform`. macOS ships bsdtar, which rejects the option and writes nothing, so the remote tar fails on the empty stream and `just` stops there. It fails loudly, but the recipe is simply unusable from a Mac.

Replaced with one tar per destination, extracted with `-C`. Files are named individually rather than by directory, deliberately: an archive built from a staged tree also carries members for `etc/` and `mnt/` themselves, and extracting those as root would reset the mode and ownership of the device's real `/etc` and `/mnt` to whatever the build host had. Because `-C` replaces `-C /`, tar can no longer create a target's parent as a side effect, so the remote `mkdir -p` now covers every `-C` target.

Verified that all 40 files land at exactly the paths the `--transform` version produced, and that no archive on the deploy path contains a directory member.

**2. A partial deploy could report success.** A local tar error — an unmatched glob, an unreadable file — still produces a valid archive that busybox tar extracts and exits 0 on, and `just`'s default linewise `sh -cu` has no pipefail. The copy step now lives in a private recipe with `set -euo pipefail`.

**3. `kill` never stopped the WAF app, so deploy wedged it.** It switches the foreground away from BTManager but leaves its `mesquite` process running. `deploy` then deletes `/var/local/mesquite/com.lzampier.btmanager` from under the live app and starts it again, which wedges the UI in a way that presents as the whole device being frozen. Two deploys in quick succession is enough.

Adding a `pkill` for it is not sufficient on its own, which is the part I'd most like you to sanity-check. All the pkills run inside one ssh command whose own shell has the patterns in its cmdline, so `pkill -9 -f "main.py --daemon"` SIGKILLs that shell and every later command in the list silently never runs. Reproduced on busybox 1.31.1 and 1.38.0: wrapper exits 137, the later pkills are unreachable. The patterns are now bracket-escaped so they cannot match the wrapper. That also revives `pkill -9 -f daemon.py`, which has been dead for the same reason.

`deploy-koreader` had the same `--transform` problem, plus a subtler one: tar matches a member name without its trailing slash, so that rule's own trailing `/` could never match the directory member, and `-C /` created a stray `/koreader-plugin/hidpassthrough.koplugin` on the root filesystem.

**Deliberately not included:** `deploy` remounts root read-write and never restores it, unlike `remove-autostart` and `install-udev`. Doing that safely means waiting for the daemon to be up first, since `server` backgrounds it and it opens `/var/log/hid_passthrough.log` on start — so it wants its own change. Happy to send that separately if you'd like it.

Tested from macOS against a PW5. I have not tested from a Linux host, though I did confirm the new archives produce identical member sets under GNU tar.

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,23 +23,7 @@ deploy:
     @echo "Remounting filesystems as writable..."
     ssh {{host}} "/usr/sbin/mntroot rw && mount -o remount,rw /mnt/base-us"
     @echo "Copying all files via tar pipe..."
-    (cd {{src_dir}} && tar cf - \
-        --transform='s|^kindle_hid_passthrough/hid-passthrough-dev.upstart|etc/upstart/hid-passthrough.conf|' \
-        --transform='s|^kindle_hid_passthrough/|mnt/us/kindle_hid_passthrough/|' \
-        --transform='s|^assets/99-hid-keyboard.rules|etc/udev/rules.d/99-hid-keyboard.rules|' \
-        --transform='s|^scripts/dev_is_keyboard.sh|mnt/us/kindle_hid_passthrough/scripts/dev_is_keyboard.sh|' \
-        --transform='s|^illusion/BTManager/|mnt/us/kindle_hid_passthrough/illusion/BTManager/|' \
-        --transform='s|^illusion/BTManager.sh|mnt/us/kindle_hid_passthrough/illusion/BTManager.sh|' \
-        kindle_hid_passthrough/*.py \
-        kindle_hid_passthrough/config.ini \
-        kindle_hid_passthrough/BUILD_SHA \
-        kindle_hid_passthrough/hid-passthrough-dev.upstart \
-        kindle_hid_passthrough/modules/*.ko \
-        assets/99-hid-keyboard.rules \
-        scripts/dev_is_keyboard.sh \
-        illusion/BTManager/* \
-        illusion/BTManager.sh \
-    ) | ssh {{host}} "tar xf - -C /"
+    @just host={{host}} push-files
     -ssh {{host}} "udevadm control --reload-rules" 2>/dev/null || true
     @echo "Clearing Python bytecode cache..."
     ssh {{host}} "rm -rf {{remote_dir}}/__pycache__"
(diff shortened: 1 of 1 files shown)


zampierilucas:
Hey @kal-kaliper, this is really good work, thanks for digging into it.

I reproduced all three. the pkill self kill is real, on busybox sh the wrapper takes the SIGKILL at the first pkill and nothing after it runs. bracket escaped, everything runs and the target still dies. and `/koreader-plugin/hidpassthrough.koplugin` is sitting on my kindle's rootfs dated April 8, so that one has been there a while. also diffed both archives, same 40 files at the same paths.

on the mesquite pattern you scoped it correctly. my device also runs `mesquite -l xyz.penguins184.kindleforge` and `mesquite -d com.lab126.stored`, so a looser match would have taken out KindleForge and stored along with it.

few things I'd change.

drop the `pkill -9 -f "daemon[.]py"` line instead of reviving it. there's no daemon.py entry point, it's an imported module and `main.py --daemon` is the only thing that runs. unanchored it matches anything with daemon.py in the cmdline, and since it has never actually executed on a device we'd be switching on a `-9` with zero test miles.

the mesquite pkill has no `-9` and nothing waits on it, so if mesquite ignores the TERM, deploy still hits `rm -rf /var/local/mesquite/com.lzampier.btmanager` under a live app, which is the wedge you're fixing. there are a handful of ssh round trips in between so it'll usually win, but a TERM then KILL would close it properly.

the mkdir comment isn't right. tar does create missing intermediate dirs for member paths, busybox included. what it won't create is the `-C` target itself. the mkdirs are needed, just for a different reason, and `{{remote_dir}}/modules` in that list is redundant since it isn't a `-C` target. worth fixing since it lives in the justfile forever.

the upstart staging can lose the mktemp and trap entirely.

    ssh {{host}} "cat > {{upstart_conf}}" < {{src_dir}}/kindle_hid_passthrough/hid-passthrough-dev.upstart

last one, `deploy-koreader` got the transform fix but not the pipefail one, it's still a plain linewise recipe, so it keeps the silent partial deploy hole you're arguing against for push-files. its archive also carries a `hidpassthrough.koplugin/` directory member, harmless since you rm -rf the target first and it's under /mnt/us, but it means the no directory members line isn't quite true across the board.

sort those out and I'll merge it. thanks again!

PR #138, opened 2026-07-30
Keep the Kindle awake while a HID device is in use

### LINKED ISSUE #136: Automatic sleep issue
Hi, @zampierilucas  
The current version 3.11 already supports key mapping, so you can ditch kindle-button-mapper, but kindle-button-mapper does support preventing sleep, and it seems kindle-hid-passthrough doesn’t have a sleep prevention feature. Could we add sleep prevention?

zampierilucas:
powerd's idle timer only counts native input, so turning pages from a Bluetooth remote leaves the Kindle looking idle and the screensaver kicks in mid-read. kindle-button-mapper has had this covered for a while, we never did.

The daemon now re-arms powerd's timer whenever a report comes in, using the same `lipc-set-prop touchScreenSaverTimeout 1` poke KOReader's AutoSuspend already uses on Kindle. Throttled to once every four minutes, which sits under powerd's 10 minute t1 timeout, and it only counts reports that actually changed so a gamepad parked at rest streaming an identical report can't pin the screen on forever. The poke runs on a throwaway thread so it never blocks the report path.

It re-arms the timer rather than setting `preventScreenSaver 1`, so the screensaver stays enabled and the power button still sleeps the device on demand. KOReader's own code notes that disabling the screensaver outright makes the t1 timeout misbehave.

Worth noting for anyone who lands here from the issue, nothing was needed on the KOReader plugin side. `autosuspend.koplugin` already makes this exact call every four minutes, and injected keys already reset its idle timer since the plugin hands the fd to `Device.input:fdopen()`. It just does nothing for people reading in the native framework, which is why this belongs in the daemon.

No config knob, it is on always.

Tested locally with a stubbed subprocess, 50 rapid reports collapse to a single poke and a second fires after the interval. Not yet verified against a live powerd, I could not reach the device. If you can give it a run, connect a remote and leave it sitting past the usual screensaver timeout while pressing a key every couple of minutes.

Fixes #136

### DIFF
--- a/kindle_hid_passthrough/host.py
+++ b/kindle_hid_passthrough/host.py
@@ -8,6 +8,7 @@ from typing import List, Optional
 from bumble.core import InvalidStateError
 from bumble.hci import HCI_LE_SET_PRIVACY_MODE_COMMAND, HCI_LE_Set_Privacy_Mode_Command, HCI_Write_Class_Of_Device_Command, HCI_Write_Local_Name_Command
 
+import keep_awake
 from ble import BLEMixin
 from bt_setup import ensure_uhid
 from classic import ClassicMixin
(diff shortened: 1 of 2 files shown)


zampierilucas:
Closing this. Digging into it properly, KOReader already does exactly this poke, it just stops doing it when auto-suspend is turned off. Moving the fix into the KOReader plugin instead of doing it behind KOReader's back from the daemon, details in #136.

PR #139, opened 2026-07-30
fix(koplugin,install): adopt non-keyboard HID devices, install from any cwd

zampierilucas:
Fixes #135.

The three symptoms in the issue are one failure plus its knock-on effects.

## The installer never installed the plugin

`installKOReaderPlugin` copied from the relative path `koreader-plugin/hidpassthrough.koplugin`, resolved against the caller's working directory rather than the extracted release. Run it as `sh /mnt/us/kindle_hid_passthrough/scripts/install.sh` from anywhere else and the copy fails, and it had already run `rm -rf` on the destination.

Reproduced on a PW12 with the v3.11.0 script:

```
$ cd / && sh /tmp/old_install.sh installKOReaderPlugin
 -> Installing KOReader plugin
 -> Ready.
exit=0
--- plugin dir after the v3.11.0 installer ---
total 16
drwxrwxrwx    2 root     root          8192 Jul 30 19:56 .
drwxrwxrwx   45 root     root          8192 Jul 30 19:56 ..
```

An empty directory, reported as success. That is symptom 1, and symptom 2 follows from it: `event_map_extra.lua` does ship in the release tarball, it just never got copied, and putting it in `koreader/settings/` does not help because the plugin loads it from its own directory.

`installUdevRules` and `installUpstart` had the same working-directory dependency. Everything now copies out of the tree the script lives in, the plugin source is checked before the destination is removed, the three plugin files are verified afterwards, and failures reach the exit status. `scripts/` and `assets/` are copied into the install dir too, since the udev rule references `dev_is_keyboard.sh` there and nothing was putting it in place for an out-of-tree install.

## Nothing ever opened the remote

This is the real cause of symptom 3. KOReader's `externalkeyboard` calls `fbink_input_check(path, INPUT_KEYBOARD, 0, 0)`, and FBInk only sets `INPUT_KEYBOARD` when *every* keycode from 1 to 31 is present (Esc, the number row, Q through D):

```c
/* the first 32 bits are ESC, numbers, and Q to D;
 * if we have all of those, consider it a full keyboard */
unsigned long mask = 0xFFFFFFFE;
if ((bitmask_key[0] & mask) == mask) {
        dev->type |= INPUT_KEYBOARD;
}
```

[...]

PR #148, opened 2026-07-31
fix(koplugin): adopt a keyboard externalkeyboard raced and dropped

zampierilucas:
Found while diagnosing why a BT keyboard cannot be mapped after a reconnect. Captured on a PW12 with debug logging on:

```
20:38:46  key event => code: 10040 (UsbDevicePlugIn), value: 3
20:38:47  [FBInk] [fbink_input_check] open `/dev/input/event3`: No such file or directory!
20:38:47  ExternalKeyboard:setupKeyboard: /dev/input/event3 doesn't look like a keyboard
20:38:48  [FBInk] /dev/input/event3: `BT Keyboard` = MOUSE|KEY|KEYBOARD|POWER_BUTTON|...
```

`externalkeyboard` schedules its check 0.5s after the uevent. On this device the `/dev/input/eventN` node does not exist yet at that point, `fbink_input_check` fails to open it, and the plugin declines the device permanently. It only retries on the next insert event.

We check at 1s. By then the node exists and FBInk reports `KEYBOARD`, so my exclude mask from #139 dropped it as well. Result: `/dev/input/event3` open by nobody.

```
$ ls -l /proc/<koreader>/fd | grep event
10 -> /dev/input/event2     # kindle-button-mapper
11 -> /dev/input/event4     # test device
9  -> /dev/input/event1     # touchscreen
                            # event3, the BT keyboard, is absent
```

That matches the symptom exactly: the keyboard types fine right after a KOReader start, because the startup scan sees a node that already exists and adopts it, and it is dead after any disconnect and reconnect.

Excluding keyboards by classification was the wrong test. The real question is ownership, and `_attachInput` already answers it one line earlier:

```lua
if Device.input.opened_devices[path] and not input_fds[path] then return end
```

So `INPUT_KEYBOARD` comes out of the mask. If `externalkeyboard` holds the device we still return early and it keeps it; if it lost the race, we take it instead of leaving it unowned. Our check runs after theirs (1s against 0.5s, and their init scan runs before ours alphabetically), so their decision is always already made when we look.

## Verified

The diagnosis is hardware-confirmed on a PW12, including that the capture dialog itself works correctly for an `externalkeyboard`-adopted keyboard (instrumented `KeyCapture:onKeyPress` fires). The fix itself is **not** yet confirmed on hardware; the device went unreachable before I could redeploy. Reproducing needs a real BT keyboard disconnect and reconnect, which the synthetic uinput devices I was testing with do not exhibit, since their nodes appear synchronously.

## Known gap

If `externalkeyboard` loses this race on a device where it holds no other keyboard, we now adopt the keyboard but its scancodes are not in `Device.input.event_map`, because `externalkeyboard` merges that map only for devices it adopted itself and a PW12 ships two page-turn codes. The keys would reach KOReader and still do nothing. That is the case the bundled keyboard map would have covered. Worth deciding separately.

### DIFF
--- a/koreader-plugin/hidpassthrough.koplugin/main.lua
+++ b/koreader-plugin/hidpassthrough.koplugin/main.lua
@@ -211,7 +211,9 @@ end
 -- Key device attach
 ------------------------------------------------------------------------------
 -- externalkeyboard matches INPUT_KEYBOARD, which FBInk only sets when keycodes
--- 1..31 are all present, so a remote or gamepad is opened by nobody.
+-- 1..31 are all present, so a remote or gamepad is opened by nobody. Keyboards
+-- are skipped by ownership below, not by type: externalkeyboard checks 0.5s
+-- after the uevent and declines outright if the node isn't there yet.
 
 -- Lazy: the fbink_input cdef above is a pcall, so C.INPUT_* at module scope
 -- would take the plugin down when it isn't there.
@@ -219,7 +221,6 @@ local exclude_types
 local function excludeTypes()
     if not exclude_types then
         exclude_types = bit.bor(
-            C.INPUT_KEYBOARD,
             C.INPUT_TOUCHSCREEN,
             C.INPUT_TABLET,
             C.INPUT_SCALED_TABLET,


zampierilucas:
Do not merge. Tested on hardware and it makes things worse.

Deployed to a PW12, then reconnected the keyboard a few times. File descriptors held by KOReader went from one per device to this:

```
13 -> /dev/input/event2
14 -> /dev/input/event3 (deleted)     <- stale handle to a destroyed node
15 -> /dev/input/event2
16 -> /dev/input/event3
17 -> /dev/input/event2
18 -> /dev/input/event3
19 -> /dev/input/event3
```

followed by:

```
21:52:45 WARN Polling for input events returned an error: 19 -> No such device
```

A stale fd in the poll set makes every poll return ENODEV immediately, which starves real input, so the keyboard stopped working entirely rather than just being unmappable.

Cause: `checkKeyDevice` calls `fbink_input_check(path, INPUT_KEY, exclude, 0)` with `settings = 0`, which keeps the fd open on a match. While keyboards were excluded they never matched, so FBInk closed the fd every time. Dropping `INPUT_KEYBOARD` from the mask made them match, and across reconnect cycles ownership alternates between this plugin and `externalkeyboard`, each of which tracks its own table and closes only what it thinks it owns. Handles accumulate.

Rolled back on the test device, which returned to one fd per device and no further ENODEV.

Worth noting separately: those ENODEV warnings predate this change. They appear at 07/30 20:36 and 07/31 20:26, 20:35 and 20:43, each lining up with a keyboard disconnect, on stock v3.12.0. A stale fd surviving a disconnect may be the actual cause of "keyboard is dead after a reconnect", independent of the exclude mask, and that is where I would look next rather than at device-type filtering.

PR #160, opened 2026-08-04
feat(install): bundle kindle-button-mapper with the release

zampierilucas:
First step of the refactor in `docs/refactor-input-ownership.md`. Gamepad and D-pad mapping belongs in kindle-button-mapper, but installing it meant a second download, so in practice nobody has it and every joystick issue lands here instead.

The release now ships the mapper under `button-mapper/` and `scripts/install.sh` runs its installer as part of `installAll`, also available as menu item 8 or `sh install.sh installButtonMapper`.

It still installs into `/mnt/us/kindle-button-mapper` with its own upstart job, so an existing standalone install is updated rather than duplicated and it keeps working if this one is removed. Uninstall leaves it alone for the same reason and prints how to remove it. The mapper version is pinned in the workflow so a release over there cannot change what this tarball ships without a commit here, and its files are in the tarball manifest check so a broken bundle fails the build.

The docs commit also corrects two blockers that turned out to be already solved. The mapper has been multi-device since 1.4.x with one worker thread per `[device.NAME]` block, and `uinput.ko` is already bundled here for the Broadcom kernels that disable it. Nothing needs to write into the mapper's config either, its WAF app already lists every `/dev/input` node with name and uniq, so a connected device is pickable without typing a MAC.

## Testing

Verified locally, `sh -n` on the installer, YAML parse on the workflow, all three `installButtonMapper` branches dry-run (not bundled, installer ok, installer fails), the exact CI download and chmod snippet against v1.4.1, and the manifest paths resolving inside the tarball. Bundled weight is about 615 KB compressed. KindleForge drives the installer by action name and not menu number, so the renumber is safe.

Not yet run on device, my Kindle was unreachable while writing this. Wants a check that `installAll` lands the mapper with its upstart job started and the MapperManager icon in the library.

## Still open

Nothing in BT Manager mentions the mapper, so a bundled tool nobody is told about is dead weight. Grab semantics are undecided too, right now the plugin and the mapper can both act on the same press.

### DIFF
--- a/.github/workflows/build-arm.yml
+++ b/.github/workflows/build-arm.yml
@@ -88,6 +88,24 @@ jobs:
           cp -r koreader-plugin/ output/
           cp kindle_hid_passthrough/config.ini output/
 
+      # Pinned rather than tracking latest, so a mapper release can't change
+      # what this tarball ships without a commit here.
+      - name: Bundle kindle-button-mapper
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          KBM_VERSION: v1.4.1
+        run: |
+          mkdir -p output/button-mapper
+          gh release download "$KBM_VERSION" \
+            --repo zampierilucas/kindle-button-mapper-rs \
+            --pattern 'kindle-button-mapper-armv7.tar.gz' \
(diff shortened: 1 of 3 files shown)


zampierilucas:

PR #161, opened 2026-08-04
feat: register paired devices with kindle-button-mapper

zampierilucas:
Second leg of the mapper-as-backend rework, after #160 ships the mapper in the release. Pairing a pad used to be half the job, the mapper only picks a device up once someone writes its MAC into its config.ini by hand or through the MapperManager app.

The daemon now appends the `[device.X]` block itself when a device is paired, from the one choke point every pair path routes through, so BTManager and the KOReader plugin both get it for free. A sweep at daemon start covers devices paired before the mapper was installed. The block carries only name and uniq, the mapper decides from the node's capabilities whether the default gamepad layout applies (zampierilucas/kindle-button-mapper-rs#37), so registering a keyboard is harmless.

Appends rather than rewrites, the config is hand-edited and a parser round-trip would eat comments. The mapper job is restarted only if already running, a stopped one reads the new block on its next start.

With #160, kindle-button-mapper-rs#37 and this, the flow becomes pair a pad, block gets written, mapper grabs it by uniq and emits arrows/Enter/page turns on its keyboard node, KOReader adopts that natively. No plugin involvement.

## Testing

6 new pytest cases cover registration, dedup against a `/P`-suffixed uniq, slug collisions, the nameless fallback and the wildcard skip. Full suite 20 passing, ruff clean.

Tested on my Basic 5. The startup sweep registered both already-paired devices as `[device.keychron_m6]` and `[device.bt_keyboard]` with correct names and uniqs, restarted the mapper, and the mapper's caps gate left both alone since neither is a gamepad. Needs #163 on devices whose python bundle lacks termios.

### DIFF
--- a/kindle_hid_passthrough/button_mapper.py
+++ b/kindle_hid_passthrough/button_mapper.py
@@ -1,0 +1,99 @@
+"""Register paired devices with kindle-button-mapper.
+
+Appends a [device.X] block with the device's MAC to the mapper's config, so
+its daemon picks the uhid node up by uniq the moment it connects. The mapper
+decides from the node's capabilities whether a default layout applies, so
+registering a keyboard is harmless.
+
+Append-only on purpose: the config is hand-edited and edited by the mapper's
+own app, a parser round-trip would eat comments and formatting.
+"""
+
+import logging
+import os
+import re
+import subprocess
+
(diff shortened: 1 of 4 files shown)


zampierilucas:

PR #162, opened 2026-08-04
feat(koplugin): edit Button Mapper mappings from inside KOReader

zampierilucas:
Third leg of the mapper-as-backend rework, after #160 (ship the mapper in the release) and #161 (register paired devices with it). The plugin already drives the hid daemon over 8321, this gives it the same role for the mapper over its helper API on 8322, so pairing and remapping live in one frontend.

A Button mappings menu lists the devices registered in the mapper's config. Map a button captures a press through `GET /capture`, which pauses the mapper's grab while it reads, then offers the mapper's own action list and writes the result back through `POST /config` and a reload. Existing mappings show per device and hold-to-remove works like the Key mappings menu.

The helper is the same process the MapperManager app spawns, started on demand with the same pidfile, so whichever frontend gets there first owns it and the other reuses it. Config edits are line-based on the ini text since the file is hand-edited and a parser round-trip would eat comments.

The existing Key mappings menu is untouched. It binds KOReader actions to whatever key events arrive, which now includes the keys the mapper emits, so the two menus compose, the mapper decides what a button becomes, Key mappings decides what the key does in KOReader.

## Testing

The pure config-text functions run under a standalone luajit harness, all passing. Both files luacheck clean, zero new warnings.

Tested on my Basic 5 through KOReader's own luajit against the live helper, 11 checks passing end to end, `ensureHelper` cold-spawned the helper, `getConfig`/`deviceBlocks` listed the registered devices, a live `capture` caught a synthesized pad's hat-up while the mapper had it grabbed, `captureTarget` resolved it to `device.X.dpad up`, and the write went through `POST /config` plus reload and fired on the next press. KOReader restarted with the plugin and loaded clean. The menu taps themselves still want a finger.

## Known rough edge

The capture request blocks the UI thread for up to 8 seconds if nothing is pressed. Same behaviour the MapperManager app has, acceptable for a settings flow, and fixable later with a background poll if it grates.

### DIFF
--- a/koreader-plugin/hidpassthrough.koplugin/main.lua
+++ b/koreader-plugin/hidpassthrough.koplugin/main.lua
@@ -563,6 +563,201 @@ function HIDPassthrough:genKeyActionMenu(id)
     return sub_items
 end
 
+------------------------------------------------------------------------------
+-- Button Mapper frontend
+------------------------------------------------------------------------------
+-- Gamepads and remotes are the mapper's job (it grabs them and emits plain
+-- keys), so their mappings are edited in its config over its helper API
+-- rather than in KOReader's event path.
+
+function HIDPassthrough:_mapper()
+    if not self._mapper_client then
+        self._mapper_client = dofile(self.path .. "/mapper.lua")
+    end
+    return self._mapper_client
+end
+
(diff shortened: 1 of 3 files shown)


zampierilucas:

PR #164, opened 2026-08-04
refactor: kindle-button-mapper becomes the input mapping backend

zampierilucas:
Consolidates the whole migration in this repo, one commit per leg. The plugin had grown a second mapping layer inside KOReader duplicating what the mapper already does, and the devices KOReader refuses to adopt are exactly the devices that need mapping. After this, kindle-hid does Bluetooth and produces the evdev node, the mapper gives buttons meaning, the plugin is the frontend for both, and each still works alone.

The mapper-side half is zampierilucas/kindle-button-mapper-rs#37 (capability-gated default gamepad layout, bare-MAC uniq matching, in-process key injection). Merge that first and tag a mapper release, then bump `KBM_VERSION` here.

The four commits:

1. **Bundle the mapper with the release.** CI pulls the pinned mapper release into `button-mapper/` and `scripts/install.sh` runs its installer as part of `installAll`, also menu item 8 / `sh install.sh installButtonMapper`. It keeps its own directory and upstart job, so a standalone install is updated rather than duplicated and uninstalling this leaves it alone.

2. **Register paired devices with the mapper.** The daemon appends the `[device.X]` block (name + uniq) at the one choke point every pair path routes through, and sweeps devices.conf at startup for devices paired before the mapper existed. The mapper decides from node capabilities whether defaults apply, so registering a keyboard is harmless.

3. **Plugin edits mapper mappings from inside KOReader.** (superseded in part by the menu unification below) The plugin already drives the hid daemon over 8321; it now drives the mapper's helper API on 8322 the same way. Button mappings menu, capture a press (pauses the grab), assign from the mapper's action list, write back through `POST /config` plus reload, hold-to-remove. The helper is the same process MapperManager spawns, shared pidfile. The existing Key mappings menu is untouched and composes with this, the mapper decides what a button becomes, Key mappings decides what the key does in KOReader.

3. **One mapping menu, backed by the mapper.** Two menus meant two mental models for the same job. Key mappings is now the only one, and the mapper stores it. Its action list carries every no-argument KOReader Dispatcher action, generated into `koreader_actions.lua` from KOReader's own `settingsList` (122 actions in 7 sections, regenerate with `scripts/gen_koreader_actions.py`), plus the mapper's native ones for what KOReader can't do. Picking one writes `koreader.sh event <Name>` into the mapper config, so the whole mapping lives in one file that still works with KOReader closed. This drops the plugin's own keymap, capture widget, `key_events` registration and Dispatcher execution, about 200 lines.

4. **Write the plugin's scope into CLAUDE.md.** Easy cases here, hard cases in the mapper.

### Breaking

Existing bindings in `hidpassthrough_keymap.lua` stop firing and have to be re-made against the device. They were keyed by KOReader key name with no device attached, so there is nothing to migrate them onto.

Resulting flow, pair a pad anywhere, block gets written, mapper grabs it by MAC and emits arrows/Enter/page turns on its keyboard node, KOReader adopts that natively, customize from either frontend.

## Testing

All on my Basic 5 (5.16.2.1.1), plus unit suites.

- Startup sweep registered both already-paired devices with correct slugs and uniqs and restarted the mapper. The caps gate left them alone, neither is a gamepad, so nothing grabbed my mouse or keyboard.
- Synthesized gamepad, default layout applied from node caps, grabbed, vkeyboard emitted hat-up→KEY_UP, hat-left→KEY_LEFT, A→KEY_ENTER, TR→page turn.
- Synthesized keyboard, parked with `not a gamepad — leaving it alone`, KOReader kept it.
- Plugin client, 11 on-device checks through KOReader's luajit against the live helper, including cold-spawning the helper and a live capture mid-grab, write-back persisted and fired on the next press.
- KOReader restarted with the plugin, loaded clean.
- pytest 20 passing, ruff clean, `sh -n` clean, luacheck zero new warnings, standalone luajit harness for the config-text functions all green, CI bundling step ran verbatim locally against the v1.4.1 tarball.

Not yet exercised, the menu taps themselves and a real BT pairing event (the sweep covers the same code path). Dev deploys on MTK also want #163, current main's daemon dies on the termios import there.

Supersedes #160, #161 and #162.


ZongbingHou:
@zampierilucas  Hi
Glad to be able to merge together. I tested this PR. 
Known issues:
1. After Kindle-button-mapper became the backend, uninstallAll in the new install doesn't include uninstallButtonMapper.
2. After pairing, the mapped buttons don't work. At this point, you need to turn off the Bluetooth device and then turn it back on to let it reconnect automatically for it to work. This issue also occurs after Koreader restarts. It seems to be caused by the program grabbing the event.
3. Some Bluetooth device MACs in the mapper's event have a "/P" suffix. While mapping works normally, if this suffix is missing in "/mnt/us/kindle-button-mapper/config.ini", the Bluetooth device won't work properly.

zampierilucas:
@ZongbingHou Yeah sorry, this is still wip, I had a local fix for 2 but forgot to push, and 1 and 3 should be fixed now

ZongbingHou:
Can you integrate plugin installation and uninstallation into the Kual entry in the new version?

ZongbingHou:
One more issue found on the KOReader plugin side:

Currently, for certain Bluetooth devices, the plugin fails to attach the generated UHID event node, resulting in KOReader displaying Not opened - key presses will be ignored in the Input Device Diagnostic menu.

Root Cause:

1、For these specific Bluetooth devices, the daemon strips the original 128-byte descriptor down to a 41-byte key-only descriptor (Stripped digitizer collection(s) (128 -> 41 bytes) in logs) and creates the UHID node (e.g., /dev/input/event9) with Vendor=0x0000, Product=0x0000.

2、When KOReader scans the inputs, checkKeyDevice(path) runs FBInkInput.fbink_input_check() against this node.

3、Because the stripped UHID node lacks standard hardware attributes, FBInk's ioctl check fails and returns nil.

4、As a result, _attachInput() exits early without calling Device.input:fdopen().
```text
2026-08-07 21:21:33,026 INFO daemon: Daemon resuming...
2026-08-07 21:21:33,028 INFO daemon: Accepting 2 device(s):
2026-08-07 21:21:33,028 INFO daemon:   - HBTR002 (41:42:05:9C:9E:8D) (ble)
2026-08-07 21:21:33,029 INFO daemon:   - HBTR002-XT (41:42:8E:74:81:14) (ble)
2026-08-07 21:21:33,056 INFO daemon: === Continuing with paired device ===
2026-08-07 21:21:33,058 INFO ble_hid: [BLE] Using existing connection from pairing
2026-08-07 21:21:33,058 INFO ble_hid: [BLE] Connection already encrypted
2026-08-07 21:21:33,168 INFO ble_hid: [BLE] Found HID service
2026-08-07 21:21:33,437 INFO ble_hid: [BLE] Found input report 1
2026-08-07 21:21:33,557 INFO ble_hid: [BLE] Found input report 2
2026-08-07 21:21:33,799 INFO ble_hid: [BLE] Found input report 3
2026-08-07 21:21:33,917 INFO ble_hid: [BLE] Found input report 4
2026-08-07 21:21:34,037 INFO ble_hid: [BLE] Found input report 5
2026-08-07 21:21:34,158 INFO ble_hid: [BLE] Found input report 6
2026-08-07 21:21:34,159 INFO uhid_handler: Stripped digitizer collection(s) (128 -> 41 bytes)
2026-08-07 21:21:34,168 INFO uhid_handler: Created UHID device: HBTR002 (vendor=0x0000, product=0x0000, rd_size=41)
2026-08-07 21:21:34,169 INFO ble_hid: UHID device created: HBTR002
2026-08-07 21:21:34,279 INFO ble_hid: [BLE] Subscribed to report 1
2026-08-07 21:21:34,429 INFO ble_hid: [BLE] Subscribed to report 2
2026-08-07 21:21:34,434 WARNING bumble.host: HCI_LE_CONNECTION_UPDATE_COMMAND failed (LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR)
2026-08-07 21:21:34,435 ERROR asyncio: Task exception was never retrieved
future: <Task finished name='Task-55' coro=<Host.send_async_command() done, defined at /mnt/us/kindle_hid_passthrough/dist/bumble/host.py:804> exception=HCI_Error()>
Traceback (most recent call last):
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/host.py", line 828, in send_async_command
bumble.hci.HCI_Error: HCI_Error(hci/LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR [0x23])
2026-08-07 21:21:34,549 INFO ble_hid: [BLE] Subscribed to report 3
2026-08-07 21:21:34,553 WARNING bumble.host: HCI_LE_CONNECTION_UPDATE_COMMAND failed (LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR)
2026-08-07 21:21:34,554 ERROR asyncio: Task exception was never retrieved
future: <Task finished name='Task-56' coro=<Host.send_async_command() done, defined at /mnt/us/kindle_hid_passthrough/dist/bumble/host.py:804> exception=HCI_Error()>
Traceback (most recent call last):
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/host.py", line 828, in send_async_command
bumble.hci.HCI_Error: HCI_Error(hci/LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR [0x23])
2026-08-07 21:21:34,669 INFO ble_hid: [BLE] Subscribed to report 4
2026-08-07 21:21:34,673 WARNING bumble.host: HCI_LE_CONNECTION_UPDATE_COMMAND failed (LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR)
2026-08-07 21:21:34,674 ERROR asyncio: Task exception was never retrieved
future: <Task finished name='Task-57' coro=<Host.send_async_command() done, defined at /mnt/us/kindle_hid_passthrough/dist/bumble/host.py:804> exception=HCI_Error()>
Traceback (most recent call last):
  File "/mnt/us/kindle_hid_passthrough/dist/bumble/host.py", line 828, in send_async_command
bumble.hci.HCI_Error: HCI_Error(hci/LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR [0x23])
2026-08-07 21:21:34,796 INFO ble_hid: [BLE] Subscribed to report 5
2026-08-07 21:21:34,915 INFO ble_hid: [BLE] Subscribed to report 6
2026-08-07 21:21:34,976 INFO ble_hid: [BLE] Protocol Mode: Boot
2026-08-07 21:21:34,977 INFO ble_hid: [BLE] Forced Report Protocol Mode
2026-08-07 21:21:34,978 INFO ble_hid: [BLE] Wrote Exit Suspend to HID Control Point
2026-08-07 21:21:34,979 INFO ble_hid:
[BLE] Paired and receiving HID reports. Press Ctrl+C to exit.

**Comment by usersmss** [comment]
Hi! I think I found an edge case/regression in v3.14.0 with a composite BLE HID page turner, and I wanted to report my findings in case they are useful.

I'm using a **Smart 1-P** BLE page turner with **kindle-hid-passthrough v3.14.0 / Button Mapper v1.4.1**.

The device exposes two separate input nodes:

```text
Smart 1-P Consumer Control
/dev/input/event3
[key, volume]

Smart 1-P Keyboard
/dev/input/event4
[key, keyboard, dpad]
```

### What happens

With the default configuration in v3.14.0:

* The device pairs and connects successfully.
* In the KOReader Key Mapping UI, the normal keyboard-mode buttons cannot be captured.
* If I switch the controller to its media/volume preset, buttons can be captured as `114/115` (`KEY_VOLUMEDOWN` / `KEY_VOLUMEUP`), but mappings still do not work correctly.
* In keyboard mode, the actual useful keys are:

  * `103` = KEY_UP
  * `108` = KEY_DOWN
  * `57` = KEY_SPACE

### Workaround that finally works

In MapperManager I had to:

1. Configure the device by the exact node name:
   `Smart 1-P Keyboard`
2. Remove/leave the uniq/MAC empty.
3. Enable Exclusive handling.
4. Capture the buttons there instead of through KOReader.
5. Map:

   * 108 → Next page (tap)
   * 103 → Previous page (tap)
   * 57 → a custom KOReader action

With this workaround, page turning now works in **both the native Kindle reader and KOReader**.

However, after doing this, KOReader's own Key Mapping screen reports:

> Smart 1-P Keyboard is not connected

even though the BLE device is clearly connected and the existing mappings are working.

So I can no longer reliably add/capture new mappings directly from KOReader. I have to go back to MapperManager and configure the keyboard sub-device manually.

This feels like it may be related to composite HID devices that expose multiple `/dev/input/event*` nodes under the same Bluetooth device (in this case **Consumer Control + Keyboard**).

Would it be possible for the KOReader/mapper integration to select the appropriate input node by **name/capabilities**, or otherwise support multiple event nodes belonging to the same HID device?

In v3.11.0, the keyboard mode worked much more directly inside KOReader: Up/Down keys could navigate the UI and mappings could be configured directly there.

### One other regression / usability issue

I also use the third-party **Bookshelf** KOReader plugin.

In v3.11.0 I could bind its shortcut directly through KOReader. In v3.14.0, the Bookshelf action does not appear in the mapping action list, so I had to manually create a custom command for:

`BookshelfOpenStartMenu`

It would be great if third-party Dispatcher actions could also appear dynamically in the KOReader mapping UI.

I'm happy to test any fix or development build on this Smart 1-P device if that would help. Thanks for maintaining this project!

zampierilucas:
> Can you integrate plugin installation and uninstallation into the Kual entry in the new version?

kual entry? we got rid of that looong ago as it was getting insane to keep all the installation methods working, but hmm maybe we can read now that we have a single install.sh script

Issue #179, opened 2026-08-13
Title: MTK PW5 firmware initialization issue (040-juno_1902 / Feb 2026 firmware) - btmanagerd collision


vitorroman17:
Hello! I'm trying to use kindle-hid-passthrough on a PW5 (detected as code 0x690) running a very recent Amazon firmware (040-juno_1902_malbec_bellatrix-467076, kernel 4.9.77 from Feb 2026).

It seems Amazon completely replaced bluetoothd and vhci_stpbt with their own btmanagerd and acsbtfd upstart jobs. When kindle-hid-passthrough runs, it tries to evict the holders of /dev/stpbt. The fuser -k command successfully kills btmanagerd, but upstart instantly respawns it.

If I manually disable btmanagerd and acsbtfd via initctl stop so they don't respawn, /dev/stpbt becomes completely free. However, closing /dev/stpbt drops the power to the MTK chip via the wmt_cdev_bt module, wiping its firmware. When kindle-hid-passthrough opens /dev/stpbt to scan, the chip powers back on but has no firmware patch loaded (as Bumble doesn't upload the MTK patch). As a result, the chip is uninitialized, returning 0 devices instantly (~11ms) on any --pair scan, and registering a default MAC 00:00:46:67:61:01 instead of the real one.

Is there a workaround to either inject the MTK firmware patch using Bumble on startup, or a proper way to hijack /dev/stpbt without the wmt_cdev_bt driver dropping power/firmware from the MTK chip?

zampierilucas:
Hey @vitorroman17, the btmanagerd race is real, released builds SIGKILL the holder and upstart just wins. #99 stops both jobs through upstart instead, try that branch rather than stopping them by hand.

Can you paste hid_passthrough.log from one `--pair` run? The `BLE scan error` and `Classic scan error` lines are what we need.

vitorroman17:
Hey @zampierilucas , thanks for looking into this!

I actually manually simulated exactly what your PR #99 does yesterday. I manually ran initctl stop acsbtfd and initctl stop btmanagerd before running --pair.

[...]

PR #181, opened 2026-08-15
fix(pair): only watch stdin for early stop when it is a tty

zampierilucas:
`--pair` registers an `add_reader` on stdin so Enter stops a scan early. When stdin isnt a tty (ssh with a command, a pipe, nohup, upstart) it sits readable at EOF, so the callback fires on the first loop iteration and sets the stop event before the scan starts.

The 10s scan window is a plain asyncio timer, `_interruptible_sleep` waits on that event, so both the BLE and Classic scans return in ~0ms with 0 devices and `--pair` spins forever on "No HID devices found. Scanning again...". No HCI error is logged because nothing on the controller side failed.

This is what #179 was reading as an uninitialized MTK chip, the 11ms inquiry came from our own timer rather than the radio.

Fix is to only watch stdin when `sys.stdin.isatty()`. Non interactive runs just scan for the full duration.

Also in here, the diagnostics dmesg grep now catches `wmt|stpbt|conninfra|mtk` and `fuser` holders print for `/dev/stpbt` too, so the next report says whether the kernel actually downloaded the BT patch on open.

`tests/test_pair_stdin.py` drives `pair_mode` with a fake non tty stdin and asserts the stop event isnt already set. It fails on the old code and passes on the new one, and the tty case still stops early.

Draft until @vitorroman17 confirms on a real terminal whether the chip is actually deaf.

Refs #179


vitorroman17:
Confirmed on a PW6 (0xC7E, FW 5.18.6, kernel 5.15.41), build ad85a99.

Same binary, same device, only stdin differs.

Non-tty (ssh with a command), 26 s wall clock:

No HID devices found. Scanning again... x11


Eleven iterations. At a real 10 s window only two fit, so the scan itself
returns immediately and what remains is the retry delay.

Real tty (ssh -tt):

00:51:22 Scanning BLE + Classic concurrently (10.0s)...
00:51:22 Starting Classic inquiry (10.0s)...
00:51:29 Classic: A4:63:A1:02:34:15/P CoD=0x0C010C (Computer) HID=False
00:51:32 Classic inquiry complete: 1 total, 0 HID
00:51:34 Scanning BLE + Classic concurrently (10.0s)...
00:51:44 Classic inquiry complete: 1 total, 0 HID


Full 10 s both times, and it discovers a device.

So the chip is not deaf. The inquiry works, the radio works, and #179 was
reading our own timer. The 0 HID is just that nothing HID was in pairing mode
at that moment; the Computer at A4:63 is a nearby machine and gets filtered.

PR #198, opened 2026-08-21
test: --jobs=1, measure LTO serial fraction (do not merge)

zampierilucas:
Measurement only.

Forces `-flto=1`, a single LTO partition, fully serial. Compare the `took N seconds` link figure against the **824.4s** baseline at `-flto=4`.

If the link is roughly the same, the work is serial and no amount of cores will help — renting hardware is wasted money. If it is ~4x longer, LTRANS parallelises well and a 16 core box should scale.

Expect 199 ccache misses, since `-flto=N` is part of the compile flags. Only the link number matters.

### DIFF
--- a/.github/Dockerfile.arm
+++ b/.github/Dockerfile.arm
@@ -78,7 +78,7 @@ RUN --mount=type=cache,target=/nuitka-cache \
     --mount=type=cache,target=/ccache \
     python3 -m nuitka \
     --mode=standalone \
-    --jobs=$(nproc) \
+    --jobs=1 \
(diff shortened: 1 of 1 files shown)


zampierilucas:
Answered. Link was 1526.3s at `-flto=1` vs 824.4s at `-flto=4`, a 1.85x speedup from 4 cores. So ~61% parallelises and ~590s is irreducibly serial. No rentable AArch32 core is faster per-core than the 3.4GHz Cobalt 100, so hardware cannot fix this.

PR #201, opened 2026-08-21
test: clang + ThinLTO (measurement, needs on-device check)

zampierilucas:
Measurement branch. Based on main so the only variable is the compiler and LTO mode — compare the link against the **824.4s** baseline.

## Why

`--jobs=1` measured the link at **1526.3s** vs **824.4s** at `--jobs=4`. That is a 1.85x speedup from 4 cores, so ~61% parallelises and **~590s is irreducibly serial**. Amdahl caps any machine at ~11 min, and every rentable AArch32 core is slower per-core than the Cobalt 100 runner. Cutting LTO input 199 -> 110 in #200 only moved the link 824s -> 795s, confirming the cost is the serial stage, not file count.

ThinLTO is the only remaining lever that attacks that stage directly.

## Changes

- `clang` added to apt
- `--clang --experimental=thin-lto` — Nuitka requires **both**; `SconsCompilerSettings.py:250-262` only emits `-flto=thin` when clang mode and the experimental tag are set, otherwise `-flto=$job_count`
- `-mword-relocations` dropped — **GCC-only, clang errors with `unknown argument`**

## Risk

Roughly 50/50 it builds. `-mword-relocations` was added in 2fdd725 for "ARM conditional branch out of range". `-mlong-calls` is the flag that actually addresses branch range and clang keeps it, but dropping the other is plausible rather than proven. There is also a known pattern of LTO plus ARM flags breaking under LLD.

**Green CI is not sufficient here.** A clang-built binary has different runtime characteristics on a 2-core Kindle than a GCC one, and runtime is the whole reason LTO stays on. Needs on-device verification before this is considered.

### DIFF
--- a/.github/Dockerfile.arm
+++ b/.github/Dockerfile.arm
@@ -15,6 +15,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     patch \
     patchelf \
     ccache \
+    clang \
     libffi-dev \
     zlib1g-dev \
     linux-libc-dev \
@@ -52,7 +53,7 @@ COPY kindle_hid_passthrough/BUILD_SHA /build/kindle_hid_passthrough/
 COPY kindle_hid_passthrough/modules/ /build/kindle_hid_passthrough/modules/
 
 # ARM-specific compiler flags to fix branch range errors
-ENV CFLAGS="-O3 -mword-relocations -mlong-calls -ffunction-sections -fdata-sections"
+ENV CFLAGS="-O3 -mlong-calls -ffunction-sections -fdata-sections"
 ENV LDFLAGS="-Wl,--gc-sections"
 
 # Build syscall wrapper for Kindle kernel compatibility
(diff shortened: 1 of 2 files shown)


zampierilucas:
Closing. ThinLTO is slower than GCC LTO for this workload, not faster.

I initially misread the Nuitka log and reported the wrong number. The 517s figure was the slowest single compile, not the link. The link is the second timing, attached to the `-o main.bin` command.

| Config | Files | Link |
|---|---|---|
| GCC `-flto=4` | 199 | 824.4s |
| clang `-flto=thin` | 200 | 1086.9s |
| GCC `-flto=4` + prune | 110 | 795.2s |
| clang `-flto=thin` + prune | 111 | 814.8s |

ThinLTO loses in both comparisons, by 262s before pruning and 20s after. It also drops `-mword-relocations` and changes the binary to a clang build, which is real runtime risk on a 2-core Kindle for no build-time gain.

PR #206, opened 2026-08-21
test: -flto-compression-level=0

zampierilucas:
Cuts LTO bytecode compression I/O. **No codegen effect** — only changes how the intermediate representation is stored on disk, so a green build plus a link delta is sufficient evidence. No device test needed.


zampierilucas:
Closing. Link 823.9s vs 795.2s baseline, **29s worse**. Disabling LTO bytecode compression means more bytes read and written, and the link is not I/O bound. Consistent with tmpfs (#207) also being worse and with `--jobs=1` showing the cost is serial CPU work.

PR #207, opened 2026-08-21
test: tmpfs for /tmp during Nuitka build

zampierilucas:
Puts GCC's LTO intermediate files in RAM instead of the overlay filesystem. Runner has 16GB. **No codegen effect.** No device test needed.


zampierilucas:
Closing. Link 834.7s vs 795.2s baseline, **40s worse**, on a warm ccache (109 hits). Moving LTO temp files to RAM changed nothing because disk was never the bottleneck. The link is CPU-bound on serial whole-program analysis.

PR #208, opened 2026-08-21
test: mold linker

zampierilucas:
Replaces GNU ld for the final link. **No codegen effect** if it works. May fail outright on ARM32 + LTO plugin, which is itself a useful result.


zampierilucas:
Closing. mold 1.10.1 armhf installs fine but the link dies in 0.03s:

```
Nuitka-Scons: Backend C linking with 110 files
collect2: error: ld returned 1 exit status
```

mold does not handle the GCC LTO plugin on ARM32. Fails fast, costs nothing, ruled out.

PR #210, opened 2026-08-21
test: -ftime-report to profile the WPA stage

zampierilucas:
Profiling only, no codegen change. Makes GCC print per-pass wall clock for the LTO link.

Everything so far has inferred why the link is slow. `--jobs=1` showed ~590s is serial, but not **which pass**. This should show whether it is inlining, IPA-CP, ICF, devirtualisation or streaming.

Read the `-ftime-report` table in the link step output, not the build total.


zampierilucas:
Closing — served its purpose. The profile it produced found the real bottleneck: 129 invocations, 1118 CPU-s, min 1.0s / median 2.8s / **max 586.7s**, with scheduling 213.7s (36%) and tree PTA 145.2s (25%) inside that one partition. That led directly to #213, which cut the link 795.2s -> 396.1s.

PR #211, opened 2026-08-21
test: -fno-ipa-icf (skip identical code folding)

zampierilucas:
ICF is documented at ~15% of WPA time. Disabling it trades binary size (~10% larger in LibreOffice's numbers) for analysis time.

Keeps LTO and all other optimisation on. Baseline to beat: link **795.2s**.


zampierilucas:
Closing. Link **847.6s** vs 795.2s — **52s worse**. ICF was not a meaningful share of the cost here; the real bottleneck is one oversized LTRANS partition (see #210).

PR #212, opened 2026-08-21
test: --param lto-max-partition=50000 (split the giant partition)

zampierilucas:
`-ftime-report` on #210 found the real bottleneck: the link runs **129 invocations, 1118 CPU-s total**, distributed **min 1.0s / median 2.8s / max 586.7s**. One LTRANS partition is ~200x the median and owns 586 of 872 wall seconds, using 966MB.

That explains why every prior attempt failed — more cores, more partitions and faster I/O cannot help when a single thread owns two thirds of the time.

This caps partition size so the balancer splits it. **Changes only how work is divided, not what optimisation runs**, so no device test needed if it works.

Baseline: link **795.2s**.


zampierilucas:
Closing. Link **820.8s** vs 795.2s, 26s worse.

Capping partition size did not split the oversized partition, which means its bulk is **one function**, not a module — GCC will not split a single function across partitions. That rules out the whole partition-tuning direction, not just this parameter.

Issue #203, opened 2026-08-21
Classic reconnect: `authenticate()` as peripheral blocks setup


vitorroman17:
Sorry for the walls of text earlier — that was on me. Keeping this one to
what's actionable.

Split out of #179, Classic only. PW6 (0xC7E), FW 5.18.6, kernel 5.15.41,
Keychron K2, on the #199 build (run 32511294144).

12 sessions in 15 minutes, 10 failed. Every failure carries `Bonding restore
failed: TimeoutError()` at 5.005 s, then both HID channels at 5.007 s each,
then reason=22. The role switch was rejected 18 times and accepted 5 times in
that same window, with both outcomes appearing on both results, so the role is
a race and predicts nothing. The bonding restore predicts everything.

When the peer pages first we stay peripheral, and in BR/EDR the central drives
security. `authenticate()` goes unanswered and sits until its cap; the
exception then skips `encrypt()`, and the host pages outward for the HID
channels and raises before the peer opens them.

Calling `encrypt()` directly instead answers in 8 ms with
`HCI_Error(hci/COMMAND_DISALLOWED_ERROR [0xC])`, which is the correct reply
from the peripheral role and is harmless: the central has already encrypted
the link.

**Fix**, both gated on `is_peripheral` in `_setup_classic_session`: skip
`authenticate()`, and wait for the peer to open the channels before paging
outward. The `switch_role` request is unchanged, so a peer that accepts it
still gets a central host and #117 is unaffected.

    bond restore failed after 8ms: COMMAND_DISALLOWED_ERROR [0xC]
    channels: peer opened them in 250ms
    RESULT OK in 302ms role=PERIPHERAL encrypted=True

First attempt, no retry loop. Same fix on a PW5 (0x690, kernel 4.9.77) took
reconnect from 64 s to 253 ms.

---

Separately, and independent of the above: exceptions are logged with `{e!r}`,
and bumble's `HCI_Error.__repr__` returns `HCI_Error()`. `str(e)` gives
`HCI_Error(hci/COMMAND_DISALLOWED_ERROR [0xC])`. The `errstr()` helper already
in `logging_utils` does this. Same for disconnect reasons — `HCI_Constant.
error_name(reason)` turns `reason=19` into
`REMOTE_USER_TERMINATED_CONNECTION_ERROR`. Everything above was only findable
after that change.

zampierilucas:
[...]

PR #117, opened 2026-07-17
fix(classic): take central role on device-initiated reconnect (#85)

zampierilucas:
## What

Fixes #85: after first pairing, the 8BitDo Zero 2 (and likely other gamepads with BlueZ-targeting firmware) could never reconnect — no input device was created and the pad looped connect/drop (reason=19) forever.

## Root cause

First pairing works because we page the device and end up **central**. On reconnect the pad pages us, and bumble's `classic_accept_any` path hard-codes `HCI_Accept_Connection_Request(role=0x01)`, leaving us **peripheral**. The pad only runs its HID state machine against a central host: as peripheral, our L2CAP channel requests were ignored (5s timeout) and the pad never opened channels itself, dropping the link at ~6s.

## Fix

Request `switch_role(Role.CENTRAL)` as soon as an incoming Classic connection lands, before authenticate/encrypt (controllers commonly refuse role switches on encrypted links). Verified by @danergo: reconnect → UHID device in ~350ms.

Supporting fixes found along the way:

- clear the stale shared disconnection event when accepting a fresh link
- proactively authenticate + encrypt on bonded reconnect instead of passively waiting
- open HID channels immediately after encryption, with real error logging instead of `except: pass`
- guard handler steps with a per-connection lost flag and cancel the previous handler task (the pad reconnects ~90ms after dropping, fast enough to resurrect half-dead handler tasks)
- treat `CONNECTION_ALREADY_EXISTS` page collisions as yielding to the incoming link, not a failure
- log timeout exceptions with `repr` (`str(asyncio.TimeoutError())` is empty)

Known cosmetic leftover: once we're central the pad opens the HID channels itself and races our host-initiated open, producing harmless bumble warnings (`invalid state`, one refused duplicate channel). Everything still connects; deliberately left as is.

PR #216, opened 2026-08-22
fix(classic): leave security to the central while we are peripheral

zampierilucas:
Fixes the Classic reconnect failure in #203. Two commits, the second driven by @vitorroman17's test of the first.

When the peer pages us it runs its own LMP security procedure, and anything we issue in that window collides with it. The first commit gated on role, but their PW6 run showed the role is a coin flip and the collision is the constant. When the switch happened to succeed, `encrypt()` still collided and the peer dropped the link, so roughly 4 reconnects in 10 kept failing.

The second commit keys on the error instead. `peer_driving` is set when the role switch or the bonding attempt fails with `LMP_ERROR_TRANSACTION_COLLISION` [0x23] or `DIFFERENT_TRANSACTION_COLLISION` [0x2A], and it gates the same two things as `is_peripheral`. Skip `authenticate()`/`encrypt()` and let the peer finish the procedure it already started, and wait up to 5 s for it to open the HID channels before paging outward. A collided procedure continues on the peer's side, so the link comes up encrypted anyway. The central path also skips a redundant `authenticate()` when `connection.authenticated` is already set, which was a silent 5 s stall.

### Why this does not regress #85

The switch request is unchanged, and a peer that runs no LMP procedure of its own never collides. The 8BitDo pages us, wins the switch, collides on nothing, and lands on the untouched central path where we drive security and page the channels, exactly as #117 built it. A device that collides on the switch by timing but still needs paging eats the 5 s wait and then gets paged as before, degraded, not broken.

Untested here, no peer on hand reproduces the collision. @vitorroman17 measured the peripheral half at 308 ms and has offered runs; what still needs sessions behind it is the switch-succeeded, encrypt-collided, recovered path specifically.

### DIFF
--- a/kindle_hid_passthrough/classic.py
+++ b/kindle_hid_passthrough/classic.py
@@ -30,6 +30,9 @@ FALLBACK_HID_DESCRIPTOR = bytes([
 ])
 
 
+CLASSIC_PEER_CHANNEL_WAIT = 5.0
+
+
 class ClassicHIDChannels:
     """HID L2CAP channel pair (control 0x11, interrupt 0x13) for one connection."""
 
@@ -183,24 +186,41 @@ class ClassicMixin:
         if old is not None:
             await self._teardown_session(old)
 
-        if connection.role != Role.CENTRAL:
+        is_peripheral = connection.role != Role.CENTRAL
+
+        if is_peripheral:
             log.info("[Classic] Requesting role switch to central...")
             try:
                 await asyncio.wait_for(connection.switch_role(Role.CENTRAL), timeout=5.0)
                 log.success("[Classic] Role switch complete, now central")
+                is_peripheral = False
             except Exception as e:
                 log.warning(f"[Classic] Role switch failed: {errstr(e)}")
 
-        if not connection.is_encrypted:
+        if is_peripheral:
+            log.info("[Classic] Still peripheral, leaving security to the central")
+        elif not connection.is_encrypted:
             log.info("[Classic] Restoring bonding (authenticate + encrypt)...")
             try:
-                await asyncio.wait_for(connection.authenticate(), timeout=5.0)
+                if not connection.authenticated:
+                    await asyncio.wait_for(connection.authenticate(), timeout=5.0)
                 await asyncio.wait_for(connection.encrypt(enable=True), timeout=5.0)
                 log.success("[Classic] Bonding restored")
             except Exception as e:
                 log.warning(f"[Classic] Bonding restore failed: {errstr(e)}")
 
         channels = session.channels
+
+        if is_peripheral and not channels.intr_channel:
+            log.info("[Classic] Waiting for the peer to open the HID channels...")
+            loop = asyncio.get_running_loop()
+            deadline = loop.time() + CLASSIC_PEER_CHANNEL_WAIT
+            while loop.time() < deadline and not channels.intr_channel:
+                await asyncio.sleep(0.05)
+            if channels.intr_channel:
+                log.success("[Classic] Peer opened the HID channels")
+            else:
+                log.info("[Classic] Peer did not open them, paging outward")
 
         if not channels.ctrl_channel:
             log.info("[Classic] Connecting to HID control channel...")
--- a/kindle_hid_passthrough/classic.py
+++ b/kindle_hid_passthrough/classic.py
@@ -6,6 +6,9 @@ import asyncio
 from bumble.core import BT_BR_EDR_TRANSPORT, BT_HUMAN_INTERFACE_DEVICE_SERVICE, InvalidStateError, TimeoutError as BumbleTimeoutError
 from bumble.hci import (
     Address,
+    HCI_DIFFERENT_TRANSACTION_COLLISION_ERROR,
+    HCI_LMP_ERROR_TRANSACTION_COLLISION_OR_LL_PROCEDURE_COLLISION_ERROR,
+    HCI_Error,
     HCI_Write_Scan_Enable_Command,
     Role,
 )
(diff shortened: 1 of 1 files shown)


vitorroman17:
Tested f9fc8b1 on a PW6 (0xC7E, FW 5.18.6, kernel 5.15.41), Keychron K2.

Two consecutive sessions, opposite outcomes:


23:54:00.770  Device connected
23:54:00.987  Role switch complete, now central
23:54:01.150  Bonding restore failed: DIFFERENT_TRANSACTION_COLLISION [0x2A]
23:54:01.358  disconnected reason=19

23:54:02.002 Device connected
23:54:02.177 Role switch failed: DIFFERENT_TRANSACTION_COLLISION [0x2A]
23:54:02.178 Still peripheral, leaving security to the central
23:54:02.283 Peer opened the HID channels
23:54:02.310 receiving HID reports 308 ms


The peripheral path works exactly as intended. The gap is the other one: when
the switch succeeds, `is_peripheral` goes false, the security block runs, and
`encrypt()` collides with the procedure the peer already started. Same
`reason=19` as before.

That path is not rare. Over 14 sessions on an earlier build the switch
succeeded 6 times and collided 8, so roughly 4 reconnects in 10 still fail,
which is what the delay looks like from the user side: the first attempt dies
and the keyboard retries.

The role is not the variable, the collision is. Keying on the error code
instead covers both paths in the same shape you already have:

COLLISION = (0x23, 0x2A)  # LMP Error / Different Transaction Collision

except Exception as e:
    log.warning(f"[Classic] Role switch failed: {errstr(e)}")
    peer_driving = isinstance(e, HCI_Error) and e.error_code in COLLISION


then gate the security block and the channel wait on `peer_driving or
is_peripheral`, and set the same flag if the bonding attempt collides. #85 is
still unaffected: a peer that runs no procedure of its own never collides, so
it keeps the central path.

I have that running here, 1 session, connected first try in 414 ms. Happy to
push it as a commit on top of yours, or to collect more sessions on either.

zampierilucas:
Hey @vitorroman17, went with your collision keying, its on the pr as cb0f404 with bumble's named constants instead of the raw codes. Grab that and give me around 10 sessions? The path I still want numbers on is switch succeeds, encrypt collides, we stand by and the peer recovers. That one has a single 414ms run behind it so far, and on the old build the peer dropped the link 200ms after that exact collision. Ill run the 8bitdo style path on my end for the no collision half.

vitorroman17:
Ran `cb0f404` on the PW6 (0xC7E, FW 5.18.6, kernel 5.15.41), K2 only in
`devices.conf`, nothing else paired, log cleared first. Seven sessions.

```
 #  switch                            bonding                           channels  outcome            ms
 1  DIFFERENT_TRANSACTION_COLLISION   stood by                          -         DROP reason=19    285
 2  DIFFERENT_TRANSACTION_COLLISION   stood by                          -         DROP reason=19    333
 3  complete                          DIFFERENT_TRANSACTION_COLLISION   opened    OK                689
 4  complete                          DIFFERENT_TRANSACTION_COLLISION   opened    OK                506
 5  ROLE_SWITCH_FAILED [0x35]         stood by                          -         DROP reason=19    375
 6  complete                          DIFFERENT_TRANSACTION_COLLISION   -         DROP reason=19    519
 7  DIFFERENT_TRANSACTION_COLLISION   stood by                          opened    OK                325
```

Three reached HID reports. But `ms` is time to the first report, and reading
only that column hides the result:

**Session 4 was killed by the peer 29.9 s after the link came up.**

```
01:41:09,165 Device connected: Keychron K2 (DC:2C:26:DF:AB:DC/P)
01:41:09,377 Role switch complete, now central
01:41:09,556 Bonding restore failed: DIFFERENT_TRANSACTION_COLLISION [0x2A]
01:41:09,556 Security collision, the peer is driving; standing by
01:41:09,659 Peer opened the HID channels
01:41:09,672 receiving HID reports
01:41:39,070 Device disconnected (reason=19 REMOTE_USER_TERMINATED_CONNECTION)
```

29.9 s from ACL establishment, no suspend, no key pressed, nothing else in the
log. That is the 30 s LMP Response Timeout: the peer tore down a link whose
security procedure never completed. Keys were flowing the whole time, so from
the user side it works and then dies half a minute later.

This is the assumption the path rests on — "a collided procedure continues on
the peer's side, so the link comes up encrypted anyway". On this peer it does
not, and nothing in the code checks. Standing by waits on
`channels.intr_channel`, never on `connection.is_encrypted`.

Session 3 is the same shape and I cannot score it: it ran 13.4 s and my own
daemon suspend ended it before the 30 s mark.

So on the path you wanted numbers for — switch succeeds, encrypt collides — it
is 0 confirmed durable connections in 3: one dropped at 519 ms, one died at the
timeout, one inconclusive.

**The one that held is the opposite shape.** Session 7 collided on the *switch*,
so `encrypt()` was never called at all, the peer finished its own procedure
undisturbed, and the link is still up. Suggests the damage is not the collision
being mishandled but `encrypt()` being issued at all while the peer is mid
procedure: surviving the exception is not enough, the collision leaves the
peer's side incomplete.

Two more things from the same log:

**Non-collision switch failure still stands down.** Session 5 is
`ROLE_SWITCH_FAILED [0x35]` — no peer procedure to yield to — and the branch
skips `authenticate()`/`encrypt()` anyway. Dead at 375 ms. Third `0x35` I have
seen on this device, so it is not a one-off.

**The channel wait spins on a dead link.** Sessions 1, 2, 5 and 6 all dropped
between 285 ms and 519 ms, inside the 5 s wait, which has no
`connection.is_connected` in its loop condition. Roughly 4.5 s of dead time
before we page outward on every one of them.

What I would try next, on top of your keying: wait on `connection.is_encrypted`
rather than on the channels, drive security ourselves if that window closes
without it, and break the wait when the link is gone. Happy to collect the same
seven-bucket table for any variant you want to push.

I have not caught the 8BitDo-style no-collision path yet — this K2 collided or
failed the switch in all seven.

zampierilucas:
Closing this one in favour of #248.

@vitorroman17 your cb0f404 table is what killed it. The switch succeeded and encrypt collided path went 0 for 3, and one of those came up, passed keys and then died at the 30 s LMP timeout with the link never encrypted. So keying on the collision was gating on the wrong thing, standing down after the collision is already too late.

#248 goes the other way, it initiates nothing for 1 s and watches whether you encrypt the link or open the HID channels first. Either one means you are driving and we stay out of it, an empty window means there was nothing to collide with so it runs the old switch and security path. Session 7 of yours is the shape that held and thats the one it generalises.

Would really appreciate a run of that one on the K2 if you are still up for it. The disconnect line now prints how long the link lived and whether it was encrypted, so a session heading for the 30 s timeout is obvious in the log instead of looking fine.

PR #217, opened 2026-08-22
feat(ble): read Bluetooth page-turner battery level (0x180F/0x2A19)

1lifedo1thing:
## Summary

Add Bluetooth page-turner battery level support for connected BLE HID
devices, using the standard Battery Service (0x180F / 0x2A19).

Many BLE page turners (e.g. "Free 2") expose battery only over GATT. The
Kindle's own UI never shows it, so once a device is connected to this
daemon there is no way to see how much charge is left. This change lets the
daemon read the standard battery characteristic and expose it through the
existing HTTP API, which KOReader plugins (e.g. Bookends `%plugin_content`)
can display on the reading screen.

## What's in the PR

- After BLE HID setup completes, discover the Battery Service (0x180F),
  subscribe to Battery Level (0x2A19) notifications, and read the current
  value once. The value is cached on the session.
- A background poller re-reads the characteristic every 5 minutes as a
  fallback, so the battery stays fresh even if the device only reports on
  change.
- `DeviceSession` exposes `battery_level` / `battery_updated`, which flow
  into `GET /status` → `connections[]` automatically.
- Notifications from vendor channels are also logged and a one-byte 0-100
  value is accepted as a battery candidate, covering devices that push
  battery over a vendor characteristic instead of the standard service.

## Notes on UUID handling

Bumble exposes 16-bit UUIDs as little-endian bytes (e.g. `b'\x0f\x18'` for
0x180F), and the objects are not plain `bytes`, so direct equality against
an int fails. `uuid_bytes()` normalizes to little-endian bytes before
comparison, which works for both int and bytes representations.

## Testing

- Device: Kindle Paperwhite 12 (KPW6, MediaTek MT8512, kernel 4.9.77-lab126)
- Page turner: "Free 2" BLE HID (has 0x180F / 0x2A19, read + notify)
- Verified on daemon v3.14.1 builds:
  - Battery read immediately after connect (`battery: 90%` right after
    "receiving HID reports"), no waiting for the first notification.
  - Live updates via 0x2A19 notifications (87 → 88 → 86 ...), also visible
    as `(notify)` log lines.
  - `GET /status` returns `battery_level` in `connections[]`.
  - Works across reconnects; last value is kept when the device sleeps.
- Classic (BR/EDR) devices are unaffected; the code paths are BLE-only.

## Limitations

- Requires the connected device to expose the standard Battery Service.
- When the device is asleep it stops pushing notifications; the cached
  value stays until it wakes up.

### DIFF
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -2,6 +2,7 @@
 """BLE HID handler mixin for HIDHost."""
 
 import asyncio
+import time
 
 from bumble.core import AdvertisingData, BT_LE_TRANSPORT, InvalidStateError
 from bumble.device import Device, Peer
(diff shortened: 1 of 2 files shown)


zampierilucas:
Superseded by #222

Issue #109, opened 2026-07-15
Traceback error in bumble when a key is hit in Debug windows


mergen3107:
I don't know if this is a critical bug, but when in Debug menu of the BT Manager WAF, I can see some errors popup as soon as I hit any button on the connected keyboard.
Keyboard itself works.
This happened on all builds I tested:
newest https://github.com/zampierilucas/kindle-hid-passthrough/releases/tag/dev-d0a2c2e
bootloop fix attempt https://github.com/zampierilucas/kindle-hid-passthrough/pull/99
and clean latest release

<img width="1860" height="2480" alt="Image" src="https://github.com/user-attachments/assets/68844e0c-e6bc-437a-ae51-0f0c1c051faa" />

PR #219, opened 2026-08-22
Drop bumble async-genexpr patch, pin Nuitka 4.2

zampierilucas:
Upstream fixed the codegen bug in https://github.com/Nuitka/Nuitka/issues/3973 and it shipped in 4.2, so we no longer need to rewrite `on_att_find_by_type_value_request` before compiling bumble. Dropped the patch and the `patch` apt package that was only there to apply it.

4.2 is not on PyPI yet so the Dockerfile pulls the GitHub release tarball. Once it lands we can go back to a plain `nuitka>=4.2`.

Compiled the minimal repro from the issue with 4.2 locally and it runs clean instead of blowing up with "Task got bad yield". Still needs a run on device with a peer doing discovery to confirm the GATT server survives a Find By Type Value request, the crash from #109.


zampierilucas:
Closing in favour of the bytecode approach.

PyPI still has **4.1.3** as latest — 4.2 is not published. So this would have the release build pulling an unreleased GitHub tarball to avoid a small local patch, which is a worse trade for the binary that actually ships.

#220 merged a Nuitka plugin that can ship any module as bytecode. Bytecode modules are never compiled to C, so they are immune to this entire class of codegen bug. Adding `bumble.gatt_server` to `BYTECODE_MODULES` drops the patch **without** pinning an unreleased Nuitka, and should cut build time further.

Still needs the same on-device check this PR called for — a peer doing discovery, confirming the GATT server survives a Find By Type Value request (#109).

PR #222, opened 2026-08-22
feat: page-turner battery level, from the daemon through to the status bar

zampierilucas:
Reworked #217 and carried it through to somewhere you can actually see it.

The daemon reads the standard Battery Service. After HID setup it looks for
0x180F on the peer, subscribes to Battery Level 0x2A19, reads once, then
re-reads every five minutes for devices that answer reads but never notify.
`battery_level` and `battery_updated` land in `GET /status` under
`connections`.

BTManager shows it on the connected device rows and in the device detail
overlay. The icon is drawn with CSS boxes rather than a glyph, since WebKit
533 on the Kindle has no emoji font to fall back on.

The KOReader plugin puts it in the status bars. KOReader has a supported way
in for this, `ReaderFooter:addAdditionalFooterContent` for the bottom bar and
`ReaderCoptListener:addAdditionalHeaderContent` for the crengine header at the
top of EPUBs, so nothing gets patched. The content function runs on every
status bar refresh and cannot block, so the level is cached and a scheduled
poll refreshes it every five minutes.

Three things are gone from #217. Its poller re-ran service discovery and
re-subscribed with a fresh lambda every pass, and since bumble keeps
subscribers in a set keyed by handle while lambdas hash by identity, another
callback stacked up every five minutes. It also subscribed to every notify
characteristic outside HID and treated any single byte between 0 and 100 as a
battery reading, which would have shown wrong numbers on devices that push key
or mode bytes on vendor channels. And `uuid_bytes` was unnecessary, bumble
already exports `GATT_BATTERY_SERVICE` and `GATT_BATTERY_LEVEL_CHARACTERISTIC`
and its UUID equality normalizes to 128 bits, which is what the rest of ble.py
has always done, so the byte compare would also have missed any device
reporting the service in full 128 bit form.

Testing on device turned up three real bugs. The read was hooked to
`_setup_ble_session`, but a device that pairs on this connection reaches HID
setup through `_continue_ble_after_pairing` and never touches that wrapper, so
the first connection after pairing silently skipped it, in #217 and in my
first pass both. The plugin poll was gated on adopting an input node, which
never fires for gamepads and gets claimed by whichever of the FileManager or
Reader instance loads first since `input_fds` is module level, so the reader
instance, the only one with status bars, never polled at all. And the footer
glyph was hardcoded to `\u{e790}`, which is the icon font's charge unknown
symbol, a battery with a question mark, so it now calls
`powerd:getBatterySymbol` the way the footer does for the Kindle's own
battery.

Deployed on a Paperwhite 12. Classic is unaffected, an Xbox controller
connects with its UHID node as before. With a faked 73 percent the app renders
in both the row and the detail overlay and the footer item shows correctly.
The real GATT read is still unverified, I have no peripheral here that
advertises 0x180F, and the crengine header is unverified because Alt Status
Bar is off on my device.

One thing worth knowing for anyone who tries this and sees nothing. A BLE-M3
page turner exposes 1800, 1801, 180A, 1812 and a vendor service FAB0 holding
one notify characteristic and one write characteristic, with no readable
battery anywhere. Devices like that cannot report a level and there is no
honest way around it without the vendor protocol.

@1lifedo1thing if you can point the Free 2 at this branch it would confirm the
part I cannot.


1lifedo1thing:
@zampierilucas Tested with the Free 2 on Paperwhite 12. Battery reads immediately after connect (84%), /status reports battery_level, the footer shows the level, and BTManager shows it in the device list. Live updates via notifications confirmed.

zampierilucas:
@1lifedo1thing Great merging this, thank you for contributing this and testing!

PR #233, opened 2026-08-23
build: shrink the release from 21.4MB to 11.5MB

zampierilucas:
Cuts the release from 21.4MB to about 11.5MB. Draft because the Nuitka side only gets proven by CI, and pairing still needs a run on real hardware.

Three things.

bumble wraps its `cryptography` import in a try/except and falls back to `bumble.crypto.builtin`, a pure python AES and P-256. Nothing else in bumble imports it, so `--nofollow-import-to=cryptography` drops the 10MB `_rust.so` (it statically links its own OpenSSL, which is why it's that big) plus `_cffi_backend.so`. Worth 5.2MB compressed and no code change.

`libstdc++.so.6` was dead weight, 601KB. I walked the NEEDED list of every ELF in the tarball and nothing links it, it was just sitting in the Dockerfile copy loop.

The tarball and the kpkg now use xz. Busybox tar on device takes `-J` and I confirmed a stdin extract end to end on a Basic 5. Timed both on that device, gzip decompresses at 15.3MB/s and xz at 8.7MB/s, so for the new payload that's 2.3s against 4.0s, once, at install. Preset 6 wants a 9MB decoder buffer and there's 121MB free, `-9` would want 65MB for another 220KB so I left it at the tar default.

Side finding while checking this, KPM builds libarchive with `zlib=disabled, lzma=enabled, zstd=enabled`. Our gzip kpkg was only extracting through libarchive's external program fallback this whole time, xz is the format it actually decodes natively.

What I want out of CI is confirmation that `_rust.so` and `_cffi_backend.so` are gone from `dist/` and that the tarball still passes the file list check. Then a device test that pairing still works on the builtin crypto, ECDH benchmarks at 33ms on my laptop so call it a few hundred ms on the Kindle, once per pair. `AddressResolver.resolve` also does one AES block per stored key per advert, but it's gated on `address.is_resolvable` and typical gamepads don't use resolvable addresses.

Renaming the asset to `.tar.xz` touches the justfile and the readme, both updated here. KindleForge is unavailable right now anyway so nothing live breaks.

### DIFF
--- a/.github/Dockerfile.arm
+++ b/.github/Dockerfile.arm
@@ -95,6 +95,7 @@ RUN --mount=type=cache,target=/nuitka-cache \
     --nofollow-import-to=bumble.transport.hci_socket \
     --nofollow-import-to=bumble.transport.pty \
     --nofollow-import-to=platformdirs \
+    --nofollow-import-to=cryptography \
     kindle_hid_passthrough/main.py
 
 # Copy the build container's dynamic linker and core glibc libs into the dist.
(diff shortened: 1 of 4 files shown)


zampierilucas:
Tested on device with the CI build, Basic 4.

Fresh Classic pair works, and a fresh BLE pair with LE Secure Connections completes in 305ms and 545ms across two runs, so the pure python P-256 is doing the real work and it's fast enough that you don't notice it. Both devices held UHID nodes at the same time and delivered reports, survived a reboot, and there's no ImportError or traceback anywhere in the log.

Release is 11314084 bytes against 21414210, so 47% off. dist on device went from about 45M down to 32.1M, and the xz extract takes 8.6s on the Kindle with 4.13s of that being CPU.

One thing that isn't this PR but bit me while testing, extracting the tarball straight over an existing install leaves the old dist/cryptography and _rust.so behind, because tar doesn't delete what the archive no longer contains. So you keep the 10MB and bumble happily keeps importing it. install.sh is fine since it wipes dist first, but the manual path is silent about it. Worth its own issue.

PR #235, opened 2026-08-23
fix(uhid): hold reports until the input node can be read

### LINKED ISSUE #185: Classic: first report after connect is dropped before the UHID node exists
Split out from #172, found while chasing the SET_PROTOCOL bug there.

The peer starts sending interrupt reports before we create its UHID node, so those reports are dropped. Reproduced on every connect with an Xbox controller on a Basic 4:

```
01:34:09,015 WARNING 98:B9:EA:01:67:68 sent a report before its UHID node existed; that input is dropped
01:34:09,166 WARNING bumble.sdp: attribute ID element is not an integer
01:34:09,189 WARNING [Classic] Using fallback descriptor
01:34:09,202 INFO    Created UHID device: Xbox Wireless Controller (rd_size=86)
01:34:09,222 INFO    first report forwarded to UHID from 98:B9:EA:01:67:68
```

About 200ms, spent mostly in the SDP query for the report descriptor. `_forward_report` no-ops while `session.uhid_device` is None.

For a gamepad this is harmless, it just keeps streaming. For a page turner that connects on a button press and sends one report per press, the dropped report *is* the press the user made, so the first press after every connect does nothing.

Options, roughly in order of laziness:

- hold the last dropped report and replay it once the node is created
- create the UHID node earlier when a cached descriptor already exists (the cache is loaded at startup, so for a known device we do not need to wait on SDP at all)

The warning line comes from #183, so any log showing this will now name it.

zampierilucas:
Fixes #185.

The peer sends reports before we create its UHID node and those get dropped. Chasing it on a Basic 4 turned up the other half, a report written after the node exists is lost too while nothing has opened the evdev node yet, since evdev only starts buffering for a reader once it attaches.

Flushing on UHID_OPEN was my first idea and it does not work on this kernel. The in-kernel keyboard handler claims anything with EV_KEY bits, so the event fires at creation with no reader attached and never fires again when a real one shows up. Measured what actually happens instead and X opens a fresh node about half a second after it appears, so reports are held for a second and written when that window closes.

Two places hold them. `host.py` keeps what arrives before the node exists and hands it over at creation, and `UHIDDevice` holds everything until the window is up. The media remote creates its own node through the same class so it gets this for free.

Checked with a synthetic peer driving the real code paths and reading the evdev node, both a press before the node exists and one right after creation reach it now, both were lost before. A reader that attaches later than a second still misses the first press, which is deliberate, replaying stale input is worse than dropping it.

### DIFF
--- a/kindle_hid_passthrough/host.py
+++ b/kindle_hid_passthrough/host.py
@@ -2,6 +2,8 @@
 """HID Host — runs BLE + Classic handlers on a single Bumble device."""
 
 import asyncio
+import time
+from collections import deque
 from dataclasses import dataclass
 from typing import List, Optional
 
@@ -24,7 +26,14 @@ from logging_utils import log
 from media_remote import MEDIA_REMOTE_COD, MediaRemote
 from pairing import create_keystore, create_pairing_config
 from transport import create_bumble_device
-from uhid_handler import Bus, UHIDDevice, descriptor_is_pointer, sanitize_digitizer
+from uhid_handler import (
+    PENDING_REPORTS,
+    REPLAY_WINDOW,
+    Bus,
+    UHIDDevice,
+    descriptor_is_pointer,
+    sanitize_digitizer,
+)
 
 __all__ = ['HIDHost']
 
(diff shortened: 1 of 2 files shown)


zampierilucas:
Closing this. Holding the reports does fix the drop, but the replay is worse than the bug, on a real connect it dumped sixteen stale gamepad reports a second late. Take the replay out and there is nothing left in the diff, every line was there to serve it.

What the digging turned up is on #185.

PR #237, opened 2026-08-23
build: shrink main.bin and prune the remaining dead weight

zampierilucas:
main.bin is 5.3MB of the 11.3MB release, so it's the only thing left that really moves the number. Draft until CI tells us what the bytecode change is actually worth, and it needs a device run before merging.

The plugin already shipped `bumble.hci` and `bumble.gatt_server` as bytecode. This widens it to the whole package, which should pull a lot out of the 11MB of `.text`. It should shorten the LTO link too, since there's less C to compile.

`bumble.crypto` deliberately stays compiled. Pairing does ECDH P-256 in pure python there now that cryptography is gone, and it lands at about 300ms on device. Interpreting that would be a regression you'd feel every time you pair.

The rest is the low risk pruning, measured by deleting each thing from the real tarball and re-compressing.

Dropping ssl and _hashlib is worth 1.38MB, because it takes libcrypto and libssl with them. Nothing imports either directly. `ssl` comes in through `http.server` to `http.client`, which wraps the import in try/except, and `hashlib` falls back to the bundled `_md5`, `_sha1`, `_sha256`, `_sha512`, `_blake2` and `_sha3` when `_hashlib` is missing. The CJK multibyte codecs are another 273KB. And the two exec wrappers are forty lines of C each that were costing 363KB apiece as static glibc, so they build with musl now.

Left alone on purpose. `unicodedata.so` is 272KB but `encodings.idna` imports it and I didn't want to chase that. The X11 stack plus mousecursor is 581KB but that's deleting the cursor overlay, not shrinking it. `bz2`, `lzma`, `pyexpat` and `_decimal` are 225KB together but I couldn't pin down what pulls `decimal`, so it isn't low risk.

Worth checking on device after CI, whether pairing is still around 300ms, and that the API server still serves over plain HTTP with ssl gone.

### DIFF
--- a/.github/Dockerfile.arm
+++ b/.github/Dockerfile.arm
@@ -99,12 +99,16 @@ RUN --mount=type=cache,target=/nuitka-cache \
     --nofollow-import-to=cryptography \
     --nofollow-import-to=ssl \
     --nofollow-import-to=_hashlib \
-    --nofollow-import-to=_codecs_cn \
-    --nofollow-import-to=_codecs_hk \
-    --nofollow-import-to=_codecs_iso2022 \
-    --nofollow-import-to=_codecs_jp \
-    --nofollow-import-to=_codecs_kr \
-    --nofollow-import-to=_codecs_tw \
+    --nofollow-import-to=encodings.euc_* \
+    --nofollow-import-to=encodings.shift_jis* \
+    --nofollow-import-to=encodings.iso2022_* \
+    --nofollow-import-to=encodings.big5* \
+    --nofollow-import-to=encodings.gb* \
+    --nofollow-import-to=encodings.cp932 \
+    --nofollow-import-to=encodings.cp949 \
+    --nofollow-import-to=encodings.cp950 \
+    --nofollow-import-to=encodings.johab \
+    --nofollow-import-to=encodings.hz \
     kindle_hid_passthrough/main.py
 
 # Copy the build container's dynamic linker and core glibc libs into the dist.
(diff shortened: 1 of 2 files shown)


zampierilucas:
Tested on device, Basic 4, running v3.15.1-27769a8.

Release is 6965120 bytes against 11313360 before, so 38% off this PR alone and 67% off where we started at 21414210. dist on device went 32.1M down to 19.9M and the xz extract dropped from 8.6s to 6.35s.

Where it came from. Shipping bumble as bytecode took main.bin's .text from 11MB to 3.1MB, dropping ssl and _hashlib took libcrypto and libssl with them for 1.38MB, musl took the two wrappers from 363KB each to 30KB and 21KB, and the CJK codecs plus unicodedata and multiprocessing were another 780KB between them.

bumble.crypto stays compiled on purpose. Fresh BLE pairs completed in 332ms on the earlier build and the Xbox controller paired fine here too, so the ECDH is still fast enough not to notice. Daemon starts clean with no ImportError, ModuleNotFoundError or NameError anywhere in the log, the API still serves plain HTTP with ssl gone, and pairing registers with button-mapper as before.

Two things I got wrong on the way, in case it saves someone time later. --nofollow-import-to doesn't reach the CJK codec extensions, and it doesn't reach the encodings wrappers either, Nuitka force-includes the whole encodings package in standalone mode. Deleting the .so files after the build is the only thing that works. And musl-gcc has no kernel headers on its include path, so linux/limits.h had to become plain limits.h, which both libcs provide for PATH_MAX.

Still on the table if we want more, kindle-button-mapper is 490KB and would likely halve with opt-level="z" and fat LTO, the X11 stack is 563KB and only exists so mousecursor can run against the bundled glibc, and libpython is 1.1MB shipped whole.

PR #241, opened 2026-08-23
feat(hid): settle controller lights on connect

### LINKED ISSUE #191: Proposal: add functionality to allow setting Joy-Con player lights (or, hidraw writes to the device go nowhere (nothing reads UHID_OUTPUT))
Hey @zampierilucas, first of all I want to thank you for this project - I was able to get it working with an old right joycon I had from my switch 1, and it works perfectly.  Super excited to have a wireless page turner for my kindle (pw5)/koreader!

I noticed that when I set it up, my joycon has little indicator lights that would "spin" after I connected.  Normally on a console they would settle to highlight whichever player number you are (player 1, player 2, etc.).  I worked with AI on a solution that (I hope) is simple enough to be helpful for you/other users of this project, but if it's too niche or you're not interested, no worries at all!

Here is the description from my agent:

> The lights spin because nothing ever tells the Joy-Con to stop — a console sends
it a small "set player lights" report, and the daemon has no way to send one. The
kernel is actually fine with it: on my Paperwhite (4.9, uhid built in) it queues a
UHID_OUTPUT event every time something writes to the device's hidraw node, but
nothing reads /dev/uhid, so the event gets dropped and the write quietly does
nothing. It still returns success, which is what makes it easy to miss. Reading
those events turned out to be small — a read_output_report() on UHIDDevice, and
the host passing each payload to the device's interrupt channel. After that,
hidraw behaves like it does on any other Linux box:

    printf '\x01\x00\x00\x01\x40\x40\x00\x01\x40\x40\x30\x0f' > /dev/hidraw0

> That sets all four player lights on my Joy-Con (R) and they stay put. The same
path would cover rumble.

> That one branch is all that's needed:
https://github.com/zampierilucas/kindle-hid-passthrough/compare/main...andy2mrqz:uhid-output
It doesn't set any lights on its own — someone still has to send the report, which
is the one-liner above (I run it from a script on each reconnect). Nothing needs
configuring, though every controller has its own bytes, so if you'd rather it
happen automatically, something like an init_report field in devices.conf would do
it — your call. There's a second branch on top adding an optional
GET /send-report endpoint that does the same send over HTTP:
https://github.com/zampierilucas/kindle-hid-passthrough/compare/main...andy2mrqz:send-report
That one only matters where hidraw can't work at all, like the 8th-10th gen
backport, so ignore it if you don't need it. Both are Classic only — BLE looks
like a few lines in _process_ble_report_char, which already reads the report type
and drops anything that isn't an input report, but I had no BLE device to test
with.

zampierilucas:
Follow up to #239, which built the pipe but set no lights.

Player lights are vendor specific. There is no HID usage a controller declares for them, the kernel maps eleven LED usages and none are player indicators, so the only generic part is the report itself. This sends one when a device connects, and again on every reconnect.

Two ways in. An `[output_reports]` section in `config.ini` maps an address to a hex report, so any controller works without a code change.

```ini
[output_reports]
98:B9:EA:01:67:68 = 010000014040000140403001
```

And Joy-Con and Pro Controller ship as a built in default, matched on the device name since Bluetooth HID gives us no vendor id. They are the one family that genuinely needs it, chasing all four lights forever until a host sends subcommand 0x30. Sony's settle on their own and only gain a color, and Xbox pads have no light report at all, so I left both out rather than write CRC32 code I have no device to test. A configured report always beats a built in one, so a wrong guess is fixable from config.

One gotcha, configparser counts `:` as a key delimiter, which cuts a MAC address key at its first colon. The parser now takes `=` alone. Nothing in `config.ini` used `:` as a delimiter, values with colons like `file:/dev/stpbt` are unaffected.

Tested on a Basic 4 over BLE. A configured rumble report fires the Xbox pad the moment it connects, and renaming it to Pro Controller picks the Nintendo report instead, which the pad correctly reports it cannot take.

```
Sent init report 030f00006464400000
Device takes no report 1, init skipped
```

I have no Joy-Con here, so the Nintendo payload is right by construction rather than by test, `01` then a packet counter, eight neutral rumble bytes, subcommand `30`, player 1. Same bytes @andy2mrqz confirmed in #191 except the last, he lit all four and this lights player 1.

Closes #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,10 +161,14 @@ tail -f /var/log/hid_passthrough.log
 Paired devices are stored in `devices.conf`:
 
 ```bash
-# Format: ADDRESS PROTOCOL [NAME]
+# Format: ADDRESS PROTOCOL [NAME] [OPTIONS]
 98:B9:EA:01:67:68/P classic Xbox Wireless Controller
 5C:2B:3E:50:4F:04/P ble BLE-M3
+E8:1A:5C:03:22:9F/P classic Joy-Con (R) lights=off
+AA:BB:CC:DD:EE:FF/P classic My Gamepad report=010000014040000140403001
 ```
+
+**Device options**: `lights=on|off` toggles a controller's indicator lights, the same setting as the Lights switch on the device page in BTManager. `report=<hex>` sends a raw output report when the device connects and on every reconnect, for controllers not handled out of the box. Joy-Con and Pro Controller need neither.
 
 **Multi-device support**: Every configured device connects and stays connected at the same time, across both protocols. Sessions are tracked per address, so a keyboard over Classic and a mouse over BLE (or several of each) work together.
 
--- a/README.md
+++ b/README.md
@@ -168,7 +168,7 @@ E8:1A:5C:03:22:9F/P classic Joy-Con (R) lights=off
 AA:BB:CC:DD:EE:FF/P classic My Gamepad report=010000014040000140403001
 ```
 
-**Device options**: `lights=on|off` toggles a controller's indicator lights, the same setting as the Lights switch on the device page in BTManager. `report=<hex>` sends a raw output report when the device connects and on every reconnect, for controllers not handled out of the box. Joy-Con and Pro Controller need neither.
+**Device options**: `lights=on|off` turns the indicator lights on a Joy-Con or Pro Controller on or off, the same setting as the Lights switch on the device page in BTManager. Those two are the only controllers it applies to, since every other family settles its lights on its own or has none a Bluetooth host can reach. `report=<hex>` sends a raw output report when the device connects and on every reconnect, for anything else that needs one; it replaces the built-in report, so `lights=` is ignored on a line that sets it.
 
 **Multi-device support**: Every configured device connects and stays connected at the same time, across both protocols. Sessions are tracked per address, so a keyboard over Classic and a mouse over BLE (or several of each) work together.
 
--- a/illusion/BTManager/index.html
+++ b/illusion/BTManager/index.html
@@ -97,6 +97,12 @@
                 <span class="detail-label">Battery</span>
                 <span id="detailBattery" class="detail-value"></span>
             </div>
+            <div id="detailLightsRow" class="detail-row" style="display:none;">
+                <span class="detail-label">Lights</span>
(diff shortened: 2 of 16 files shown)


zampierilucas:
Added the UI, so turning the lights off no longer means hand editing config.ini. BTManager's device page carries a Lights toggle, backed by a `/lights` endpoint and remembered in the device cache so it survives reconnects.

The toggle only shows once the light report has actually reached the device, not merely when the name matches the table. A control that silently does nothing is worse than no control.

While testing I checked whether Xbox pads could be covered too, and they cannot over Bluetooth. The guide button LED is GIP command 0x0A over USB or the Xbox Wireless dongle, and xbledctl says outright that it is not controllable over Bluetooth at the firmware level. My pad's own descriptor agrees, input report 0x01 and a force feedback output report 0x03, no LED usage anywhere. The setting does not persist in the controller either, it resets on every power cycle, which is why xbledctl runs as a service. So dimming it on a console would not carry over to the Kindle regardless.

Still no Joy-Con here, so the Nintendo payload and the toggle rendering are right by construction rather than by test. The endpoint, the persistence and the transport are verified on a Basic 4 over BLE.

zampierilucas:
Hey @andy2mrqz if you have the time, I would really appreciate if you can test this as I dont have any joycon

andy2mrqz:
> Hey @andy2mrqz if you have the time, I would really appreciate if you can test this as I dont have any joycon

I'll test it tonight, thanks for working on this!  I'll reply back with how it goes

andy2mrqz:
Hey @zampierilucas , it worked!

I tested this on my Paperwhite 5 with a right Joy-Con and a Nintendo Pro Controller, both over Classic Bluetooth.

When each controller connected, the lights correctly stopped chasing and settled on player 1. That worked on both.  And stuff like page turns still worked on both after that.

In the "BT Manager" app, the "Lights" configuration switch was present. Turning it off darkened the lights, and turning it back on brought player 1 back. That worked too.

After disconnecting the Joy-Con and connecting again, it settled on player 1 by itself. So reconnect remembered the setting as well.

lmk if that helps or if you need any other testing!

zampierilucas:
@andy2mrqz that is perfect, thank you so much, I'll merge this and it will be part of the next release

PR #247, opened 2026-08-29
Feat/A2DP Audio Passthrough

vitorroman17:
# A2DP audio passthrough

The Kindle becomes an A2DP source: Bluetooth headphones play whatever goes
through GStreamer — KOReader TTS, audiobook plugins, anything else.

## How

**1. LIPC mock.** `audiomgrd` is stopped; a LuaJIT script (KOReader's `luajit`)
registers `com.lab126.audiomgrd` and answers `audioOutputConnected = 1`, so
applications see an output.

**2. GStreamer wrapper.** While the switch is on, `/usr/bin/gst-launch-0.10` is
a shell wrapper, the stock ELF kept as `.real`. Pipelines containing
`mixersink` are rewritten to `filesink location=/tmp/kindle_audio.fifo`;
everything else delegates untouched. The wrapper writes the caps it saw to
`/tmp/kindle_audio.fmt` — that is how the daemon learns the source rate and
channels without a control channel, and how it follows a change mid-stream.

**3. Daemon.** Reads the FIFO, resamples to 44100 Hz stereo, encodes SBC
through `libsbc`, streams over AVDTP with Bumble's `MediaPacketPump`.

AVRCP media keys arrive as a Consumer Control HID report on the existing UHID
path; audio devices have no report map, so a 33-byte descriptor is injected.
The daemon does not act on them itself — an unmapped button does nothing.

## Pause

`/media/toggle`, `/media/play`, `/media/pause`, wrapped by
`assets/audio-hack/media.sh` and listed in the plugin's mapping picker.

Pause is backpressure, not a signal: the daemon stops draining the FIFO, the
writer blocks in `write()`, playback freezes at the sample. Nothing is
discarded and the producer needs no cooperation — verified against KOReader's
audiobook plugin, whose own progress counter froze and resumed at the same
offset.

## Opt-in

Off by default, its own switch beside the HID state, exposed at `/audio` and in
the plugin menu. **Installing changes nothing under `/usr/bin`** — the switch
owns the binary swap in both directions. A device where this does not work
keeps its own audio and still runs the HID host.

The wrapper redirects every `mixersink` pipeline unconditionally, so it must
not outlive the switch. Stock `audiomgrd` and the stock binary come back on
stop and on every failure path, and the start script aborts before touching
audio if `luajit` or the wrapper is missing.

## Known limits

With the switch on and no headphones connected there is no reader on the FIFO,
so local playback blocks until one connects. Turning the switch off restores
the stock binary and unblocks it.

SBC bitpool is fixed at 35, not negotiated against the sink's configured
maximum. `audioop` is gone in Python 3.13 — the resampler needs replacing
before the build image moves that far.

## Where to look

```
NEW  kindle_hid_passthrough/audio_pipe.py  451  FIFO, resampler, SBC encoder
NEW  assets/audio-hack/                    762  mock, wrapper, start/stop, media.sh

     host.py                               274  AVDTP session, AVRCP -> UHID, descriptor
     controller.py                         203  audio switch, bypass lifecycle
     classic.py                            179  audio connect, stale link key
     main.lua                               89  audio switch, Audio actions
     api_server.py                          54  /media/*, /audio
     install.sh                              8  restore the stock binary on uninstall
```

`libsbc-dev` in the ARM image, `libsbc.so.1` bundled and required by the
tarball check so a release cannot ship with the audio path dead.

## Tested

Kindle PW6SE (5.15.41-lab126) and Kindle PW5SE, clean install, Redmi AirDots: pairing, streaming,
and pause on a headset button with no manual step — an audio device gets the
grab/passthrough mapper mode automatically. Also run on an older-generation
Kindle.

### DIFF
--- a/.github/Dockerfile.arm
+++ b/.github/Dockerfile.arm
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     linux-libc-dev \
     libssl-dev \
     libx11-dev \
+    libsbc-dev \
     rustc \
     cargo \
     && rm -rf /var/lib/apt/lists/*
@@ -103,7 +104,7 @@ RUN --mount=type=cache,target=/nuitka-cache \
 RUN cp /lib/ld-linux-armhf.so.3 /build/nuitka-out/main.dist/ld-linux-armhf.so.3 && \
     for lib in libc.so.6 libm.so.6 libpthread.so.0 libdl.so.2 librt.so.1 \
                libutil.so.1 libresolv.so.2 libnss_dns.so.2 libnss_files.so.2 \
-               libz.so.1 libstdc++.so.6; do \
+               libz.so.1 libstdc++.so.6 libsbc.so.1; do \
         src=$(find /lib /usr/lib -name "$lib" 2>/dev/null | head -1); \
         if [ -n "$src" ]; then \
             cp -L "$src" /build/nuitka-out/main.dist/"$lib"; \
--- a/.github/workflows/build-arm.yml
+++ b/.github/workflows/build-arm.yml
@@ -153,6 +153,10 @@ jobs:
             kindle-hid-passthrough \
             assets/hid-passthrough.upstart \
             assets/99-hid-keyboard.rules \
+            assets/audio-hack/start_audio_hack.sh \
+            assets/audio-hack/stop_audio_hack.sh \
+            assets/audio-hack/gst-launch-wrapper.sh \
+            assets/audio-hack/lipc_audio_mock.lua \
             scripts/dev_is_keyboard.sh \
             illusion/BTManager.sh \
             koreader-plugin/hidpassthrough.koplugin/main.lua \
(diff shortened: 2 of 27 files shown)


zampierilucas:
I new some crazy bastard would contribute this at some point :P  thank you so much, let me once it is in state you feel its mergeable, in the meantime I'll test on both my kindles

vitorroman17:
I'm testing and until today, I haven't seen any more problems.

zampierilucas:
@vitorroman17 the audio itself is good, impressed you got SBC over AVDTP
working on this thing.

But this PR is doing five different things and four of them change behaviour
for devices that have nothing to do with audio. It rewrites the classic HID
connect path, changes who holds the evdev grab, adds reboot-persisted on/off
state, and ships an uninstall script nothing calls. None of that is explained
anywhere in the description.

The connect path is the real problem. It reverts #248 and takes the replay test
from clean to eight failures, and I only found that out because I went and ran
it myself. That's a regression for every keyboard and page turner in the
project, shipped inside an audio PR, unmentioned.

I'm not merging changes to core paths that I have to reverse engineer. If a
change is needed for audio, say so in the commit and show me it doesn't break
HID. If it isn't needed for audio, it doesn't belong here.

So audio only in this one. Anything touching classic.py or the plugin's input
handling goes in its own PR with its own reasoning and I'll review it there.

PR #248, opened 2026-08-29
fix(classic): give an incoming peer a quiet window before we touch the link

zampierilucas:
Supersedes #216, same bug (#203), wider fix.

@vitorroman17's seven sessions on cb0f404 killed the premise that PR rests on. Three of the failures never reached `encrypt()` at all, they asked for the role switch, stood down, and the peer dropped the link inside 375 ms. Session 4 came up, delivered keys, then got torn down 29.9 s later on the 30 s LMP Response Timeout, so a collided procedure does not reliably continue on the peer's side the way I claimed it does. The one session that held collided on the *switch* and so never called `encrypt()`.

So the collision keying guards the wrong half. Everything we initiate lands in the peer's procedure, the switch request included, and gating on the error code only tells us we already collided.

This does it the other way round. On an incoming connection we initiate nothing for up to 1 s and just watch. If the peer encrypts the link or opens the HID channels in that window it is driving, so we stay peripheral, skip the switch and the security block, and give it the same 5 s to open the channels before paging outward. If the window closes quiet there is no procedure to collide with, and we run the old path unchanged, switch to central, authenticate, encrypt, page the channels.

Ends up smaller than #216. `peer_driving`, `_is_collision` and the collision error table all go away, and the separate channel wait folds into the same gate.

#85 is unaffected. The 8BitDo runs nothing of its own so it never signals, the window closes, and it lands on the untouched central path. It pays the 1 s, which is well inside the ~6 s it waits before giving up. The `ROLE_SWITCH_FAILED [0x35]` case from session 5 now drives security itself instead of standing down for a procedure that was never running.

Two small things riding along. A redundant `authenticate()` when the link is already authenticated was a silent 5 s stall, and there is now a warning at finalize when the link is not encrypted, since that is exactly what session 4 looked like on its way to the timeout.

Still untested on hardware, nothing here reproduces the collision. @vitorroman17 if you are up for another round, what I need this time is sessions held 60 s before scoring rather than 15, plus `is_encrypted` logged at finalize and again at 35 s. Session 3 only ran 13.4 s so it never got to prove anything either way. An HCI snoop log would also settle whether it is the switch or the link key path that actually angers the K2.

### DIFF
--- a/justfile
+++ b/justfile
@@ -151,6 +151,7 @@ test:
     python3 {{src_dir}}/tests/test_hci_parser.py
     python3 {{src_dir}}/tests/test_cpu_latency.py
     python3 {{src_dir}}/tests/test_power_policy.py
+    python3 {{src_dir}}/tests/test_classic_reconnect.py
 
 # Run mock API server for local WAF app testing
 mock-server:
--- a/kindle_hid_passthrough/classic.py
+++ b/kindle_hid_passthrough/classic.py
@@ -20,6 +20,12 @@ from logging_utils import errstr, log
 # HIDP DATA header, OUTPUT report type.
 HIDP_DATA_OUTPUT = 0xA2
 
+# An incoming peer that runs its own security procedure signals it by
+# encrypting the link or opening the HID channels, both well inside 1 s on
+# the peers measured so far.
+CLASSIC_PEER_WINDOW = 1.0
+CLASSIC_PEER_CHANNEL_WAIT = 5.0
+
 FALLBACK_HID_DESCRIPTOR = bytes([
     0x05, 0x01, 0x09, 0x05, 0xa1, 0x01, 0x85, 0x01,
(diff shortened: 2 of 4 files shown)


zampierilucas:
Supersedes #216, now closed.

@vitorroman17s seven session table over there is the reason. Gating on the collision error code was not enough, the damage is issuing anything at all while the peer is mid procedure, and the one session that held was the one where encrypt() never got called.

Also adds tests/test_classic_reconnect.py, which replays the peers side of those seven sessions against the real setup path with the same milliseconds from the table, plus the 8BitDo shape from #85 that has never had hardware coverage. Run it against cb0f404 and it shows the switch going into four links that were already dying and three sessions declaring HID ready on an unencrypted link. Read the pass criteria for what they are though, they encode this branchs design so its a regression guard, not a neutral referee, and a dropped outcome only means we sent nothing before the peer killed the link.

Still zero hardware sessions on this one.

PR #257, opened 2026-09-05
feat(koplugin): serve KOReader events on a loopback port

zampierilucas:
Mapped KOReader actions needed the HTTP Inspector plugin enabled, an extra setup step that also binds a debug API on every interface and pokes holes in the firewall. The plugin now serves the inspector's own /koreader/event/<Name> URI shape itself, loopback only on 127.0.0.1:8323, and dispatches the events internally through UIManager. Same coercion rules as the inspector's parser, so the mapper's whole event surface (page turns, dispatcher actions, kolayout's quoted TextInput) works unchanged.

A registered ZMQ caps UIManager's idle wait at 50ms, so the endpoint only runs while the mapper config actually has something aimed at KOReader (koreader.sh, auto.sh or a keyboard_layout), re-checked after every mapping edit and on resume.

Tested on the Basic with the inspector off. curl to 8323 opens the menu and types text, bad requests get a 400, and the listener comes up only when the gate matches.

Pairs with zampierilucas/kindle-button-mapper-rs PR that makes the mapper try 8323 first with an 8080 fallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ The **Start on boot** toggle at the bottom installs or removes the upstart job. 
 
 If you use KOReader, a bundled plugin gives you the same scan / pair / connect / disconnect / logs / cache controls from inside KOReader — no need to exit. Open via **cog icon (Settings) → Network → BT Manager - HID Passthrough**.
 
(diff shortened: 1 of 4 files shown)


zampierilucas:
mapper side is in zampierilucas/kindle-button-mapper-rs#71

PR #261, opened 2026-09-07
fix(ble): give the LE radio a duty cycle so Wi-Fi can share the MediaTek antenna

### LINKED ISSUE #88: [bug report] Wifi won't reconnect once it's toggled off while kindle-hid-passthrough is active
Kindle-hid-passthrough works well on my device. It's just that every time after I turn off wifi while bluetooth is on in BT manager or the kindle-hid-passthrough plugin active in koreader, it won't reconnect unless I restart my device.  Currently, the only way to avoid this is to turn off bluetooth on BT manager or turn off the plugin in koreader before enabling or disabling wifi. I need to frequently turn on/off wifi because I use an AI plugin for translation and sentence analysis and I want to disable wifi after checking a sentence to save power. It'll be great if Wi-Fi states can be managed properly while the plugin is active.

Thank you again for this amazing project!

zampierilucas:
Closes #88

On MTK kindles wifi and bluetooth share one antenna, and we never let the radio rest. Every LE scan ran with the window equal to the interval, which means receive continuously, and the accept list loop retried back to back with no sleep. Wifi that is already associated survives that, a fresh association does not, which is why it only breaks after you toggle wifi off.

Scan geometry is now 60ms in 320ms, so the radio is idle four fifths of the time, and the retry loop sleeps between attempts. Third commit is unrelated but same family, we were tearing down and reopening `/dev/stpbt` every 65s whenever the remote was switched off.

Not tested on hardware, I found this reading the code, so it needs someone with the actual device.

@blue-625 @babblingcloud @ZongbingHou if you have a few minutes, grab the build from the CI run on this PR, then with your remote switched off toggle wifi off and on a few times and see if it comes back without a reboot. Also let me know if reconnect got slower, I can tighten the 320ms if it did.

If it still wedges, this is what I need.

```
dmesg | grep -iE "wmt|stp|conninfra|connfem|wlan|wifi" | tail -60
```

### DIFF
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -109,6 +109,7 @@ class BLEMixin:
                             config.connect_timeout, peer=target_address)
 
                 if connection is None:
+                    await asyncio.sleep(self.ACTIVE_RETRY_INTERVAL)
                     continue
 
                 self._admit_ble_connection(connection, matched_dev, match_kind)
--- a/kindle_hid_passthrough/ble.py
+++ b/kindle_hid_passthrough/ble.py
@@ -36,6 +36,9 @@ HID_REPORT_TYPE_OUTPUT = 2
 # Re-read the battery level this often, for devices that never notify.
 BATTERY_POLL_INTERVAL = 300
 
+LE_SCAN_WINDOW_MS = 60
+LE_SCAN_INTERVAL_MS = 320
+
 
 class BLEMixin:
     """BLE methods for HIDHost."""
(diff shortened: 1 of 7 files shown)


blue-625:
Hey, just tested it! I toggled WiFi off/on about 4–5 times with 30‑second and 10‑minute intervals, on both the stock Kindle OS and KOReader. Reconnection is super snappy – around 2 seconds on stock, and 2–3.5 seconds on KOReader. No freezes or hangs at all. Thanks for the update, it works great! 👍

blue-625:
After I gave you that feedback earlier today, I used the device for the whole day. Now I'm running into the WiFi issue again – maybe my earlier test just wasn't long enough. I suddenly couldn't connect, so I did another test: I toggled WiFi on and off five times in a row, and it still wouldn't connect. I had to restart the Kindle. Next time, I'll try testing with different durations to see if the length of usage matters. Thanks again for your update – hope this helps!

blue-625:
Thank you for your efforts in fixing the WiFi. I downloaded today's version, but this time I soon found that I couldn't turn on the WiFi.

PR #266, opened 2026-09-17
Fix add repo command in installation

5kxh2dxqmd-afk:
Very simple fix @zampierilucas, KPM needs the protocol before the URL or your KPM repo will fail to add.


zampierilucas:
Ah nice, thank you  i was obsessed in making this url smaller so users can type less on kindle that i forgot about that.

